PR #32, opened 2018-08-09
Save default branch at checkout

nwalters512:
I couldn't do this directly in `getCandidateRepos`, as that isn't called when `--repos` are specified. I opted to add a new adapter method that can do this work for all checked-out repos after they've been checked out.

I'm beginning to think that `BaseAdapter` should actually be an abstract class instead of an interface - this new function should probably be a no-op by default.

### DIFF
--- a/src/adapters/git.ts
+++ b/src/adapters/git.ts
@@ -25,6 +25,8 @@ abstract class GitAdapter implements IRepoAdapter {
 
   public abstract getDataDir(repo: IRepo): string;
 
+  public abstract mapRepoAfterCheckout(repo: Readonly<IRepo>): Promise<IRepo>;
+
   public async checkoutRepo(repo: IRepo): Promise<void> {
     const repoPath = this.getRepositoryUrl(repo);
     const localPath = this.getRepoDir(repo);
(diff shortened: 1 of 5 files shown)


src/adapters/git.ts
```diff
@@ -33,7 +35,9 @@ abstract class GitAdapter implements IRepoAdapter {
       // Repo already exists; just fetch
       await this.git(repo).fetch('origin');
     } else {
-      await simpleGit().clone(repoPath, localPath);
+      const git = simpleGit();
+      git.silent(true);
```
franjohn21: Why did we remove logging here?

src/adapters/github.ts
```diff
@@ -70,12 +83,7 @@ class GithubAdapter extends GitAdapter {
 
   public async createPullRequest(repo: IRepo, message: string): Promise<void> {
     const { migration: { spec } } = this.migrationContext;
-    const { owner, name } = repo;
```
franjohn21: 👍 Nice that we end up saving another API request

src/adapters/git.ts
nwalters512: No logging is actually the default, see the `git()` method in this class. We disable logging because we'll print errors if they occur, they'll be duplicated if we also allow the git package to log them. We can't use `this.git(...)` here because the directory doesn't exist yet. This just makes it consistent with other usages.

PR #34, opened 2018-08-09
[Migration] Lodash 3->4

nwalters512:
This PR implements a migration from Lodash 3 to 4 using [lodash-codemods](https://github.com/jfmengels/lodash-codemods). It also standardizes imports to look like

```
import { get, set } from 'lodash`;
```

I needed a `--no-reset-on-error` flag to test this, so I implemented that in this branch. Also contains a few minor improvements to Shepherd itself.


evanhobbs:
We generally do
```js
import _ from 'lodash'

_.get(x, 'somekey');
```
and rely on the babel lodash plugin to take care of this for us. What's the thinking behind enforcing this switch as part of the migration?

examples/lodash/should_migrate.js
```diff
@@ -0,0 +1,48 @@
...
+      } else {
+        throw new Error(`Unknown package version: ${pkgName} ${version}`);
+      }
+    }
+  })
+}
+
+if (!version3deps.length && !version4deps.length) {
+  console.error('This repo does not depend on Lodash');
+  process.exit(1);
```
parshap: Aside: I wonder if there's a reasonable way for shepherd to distinguish between a real error in this script and just exiting 1 to indicate *should not migrate*.

src/commands/apply.ts
```diff
@@ -54,13 +54,17 @@ export default async (context: IMigrationContext): Promise<void> => {
...
+      const spinner = logger.spinner('Resetting repo');
+      try {
+        await adapter.resetRepo(repo);
+        spinner.succeed('Successfully reset repo');
+      } catch (e) {
+        logger.error(e);
+        spinner.fail('Failed to reset repo');
+      }
+    } else {
+      logger.info('Not resetting repo');
```
parshap: 👍

src/commands/checkout.ts
```diff
@@ -46,6 +46,9 @@ export default async (context: IMigrationContext) => {
       return;
     }
 
+    // We need to create the data directory before running should_migrate
+    await fs.mkdirsAsync(adapter.getDataDir(repo));
```
parshap: 👍

nwalters512:
I was just about to add a README explaining this. Tl;dr`lodash-codemods` can only operate on usages like `_.xxx(...)`, so we have to normalize to that style first, and then we have to normalize to something else after the fact. @parshap and I made the call that we would use the destructuring import. If you'd rather handle this migration on your own, let us know and we can opt you out of this.

nwalters512:
Just found a bug in one of the normalization transforms:

```js
import { map, identity } from 'lodash';
map([...], identity);
```

will be incorrectly transformed to

```js
import _ from 'lodash';
_.map([...], identity);
```

It should be

```js
import _ from 'lodash';
_.map([...], _.identity);
```

Gotta figure out how to look up scoping information on identifiers with recast.

PR #36, opened 2018-08-13
Use the public npm registry

nwalters512:
This regenerates our `package-lock.json` to use the public npm registry instead of Artifactory. Also adds an `.npmrc` that sets the registry to the public one to work around global overrides for NW employees.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+registry=https://registry.npmjs.org/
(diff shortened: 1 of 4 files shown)


franjohn21:
The npm pacakge name is `@nerdwallet/shepherd2` -- should we change this to `@nerdwallet/shepherd`?

parshap:
And yes, we should change the name to just shepherd.

PR #37, opened 2018-08-15
Add advanced safety features for applying and pushing

nwalters512:
Adds some improved safety checks to make sure we don't overwrite human commits, push to closed PRs, and more.

This 100% needs comprehensive documentation, but I'm going to save that for another PR where I'll do a revamp of the docs to make it more palatable to complete newcomers.

### DIFF
--- a/src/adapters/github.ts
+++ b/src/adapters/github.ts
@@ -9,6 +9,12 @@ import { paginateSearch } from '../util/octokit';
 import { IRepo } from './base';
 import GitAdapter from './git';
 
+enum SafetyStatus {
+  Success,
+  PullRequestExisted,
+  NonShepherdCommits,
+}
+
 class GithubAdapter extends GitAdapter {
   private octokit: Octokit;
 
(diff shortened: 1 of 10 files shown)


src/adapters/github.ts
```diff
@@ -74,13 +79,51 @@ class GithubAdapter extends GitAdapter {
...
+      if (safetyStatus === SafetyStatus.PullRequestExisted) {
+        throw new Error('Remote branch did not exist, but a pull request does or did; try with --force-reset-branch?');
+      } else if (safetyStatus === SafetyStatus.NonShepherdCommits) {
+        throw new Error('Found non-Shepherd commits on remote branch; try with --force-reset-branch?');
+      }
+    }
+
+    // If we got this far, we can go ahead and reset to the default branch
+    await this.git(repo).reset(['--hard', `origin/${defaultBranch}`]);
+  }
```
parshap: I don't know if it's because I'm familiar with this flow now after the past few discussions, but this logic was a breeze to read through and understand! 👍

src/adapters/github.ts
```diff
@@ -74,13 +79,51 @@ class GithubAdapter extends GitAdapter {
...
+
+  public async pushRepo(repo: IRepo, force: boolean): Promise<void> {
+    const shouldForce = false;
```
parshap: Nit: Slightly odd use of a `const` here -- might as well replace `shouldForce` with a `false` literal down below?

parshap: Actually, or do you mean set set this to `true` after a successful safety status check?

src/cli.ts
```diff
@@ -96,14 +96,19 @@ addCommand('checkout', 'Check out any repositories that are candidates for a giv
 
 const applyCommand = buildCommand('apply', 'Apply a migration to all checked out repositories');
 addReposOption(applyCommand);
-// Commander's API is weird: this will get mapped to the option `resetOnError`, which will be false if this option
-// is specified. We want to default `resetOnError` to true here.
-applyCommand.option('--no-reset-on-error', 'Keep changes in the working tree even if the migration fails', true);
+applyCommand.option('--skip-reset-branch', 'Don\'t reset branch before applying', false);
+applyCommand.option('--force-reset-branch', 'Force a reset of the branch before applying', true);
+applyCommand.option('--skip-reset-on-error', 'Keep changes in the working tree even if the migration fails', false);
```
parshap: Some documentation around the default behavior and how these flags change that and the use cases for them will be nice to have in your future documentation pass.

src/cli.ts
```diff
@@ -96,14 +96,19 @@ addCommand('checkout', 'Check out any repositories that are candidates for a giv
 
 const applyCommand = buildCommand('apply', 'Apply a migration to all checked out repositories');
 addReposOption(applyCommand);
-// Commander's API is weird: this will get mapped to the option `resetOnError`, which will be false if this option
-// is specified. We want to default `resetOnError` to true here.
-applyCommand.option('--no-reset-on-error', 'Keep changes in the working tree even if the migration fails', true);
+applyCommand.option('--skip-reset-branch', 'Don\'t reset branch before applying', false);
+applyCommand.option('--force-reset-branch', 'Force a reset of the branch before applying', true);
```
parshap: Maybe we should drop `branch` from these flags and just call them `--skip-reset` and `--force-reset`?

src/adapters/github.ts
nwalters512: ...oops, this is why we do code reviews :joy:

src/cli.ts
nwalters512: I'm trying to differentiate between the `reset` that removes uncommitted changes, the `reset` that resets the branch back to `master`, and the final `reset` that happens by default if the migration fails. That's definitely an overloaded term and I'm hoping to make the reason for the name clear in the docs.

PR #40, opened 2018-08-21
Add automatic npm deployment

nwalters512:
This will automatically release Shepherd on npm when a new tag is pushed to the repo. This is untested but will only run on master, so it might take a bit of trial and error to get things working correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,31 +2,49 @@
   "name": "@nerdwallet/shepherd",
   "version": "1.0.0",
   "description": "A utility for applying code changes across many repositories",
-  "scripts": {
-    "build": "tsc",
-    "build:watch": "npm run build -- --watch",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,31 +2,49 @@
...
+      "json",
+      "node"
+    ],
+    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(jsx?|tsx?)$",
+    "transform": {
+      "^.+\\.tsx?$": "ts-jest"
+    }
+  },
+  "files": [
+    "lib"
```
parshap: What about package.json? Any other files that may be needed?

nwalters512: That, `LICENSE`, and `README.md` are automatically included by `npm`, so we should be good!

PR #50, opened 2018-08-28
Fix commit check logic to compare HEAD

parshap:
When looking for non-Shepherd commits that may have been added to the PR branch, we should compare the current repo HEAD to the upstream branch, not `master` to the upstream branch.

@NerdWalletOSS/shepherd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ When new features, bug fixes, and so on are added to Shepherd, there should be a
 
 ## [Upcoming]
 
+* Fix commit check logic when checking for non-Shepherd commits
+
 ## v1.0.0
 
 * Initial public releast to npm :tada:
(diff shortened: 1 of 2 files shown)


parshap:
Ran into this bug when some commits were pushed to `origin/master` a few days after I ran `apply` and my local branch did not have those commits.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ When new features, bug fixes, and so on are added to Shepherd, there should be a
 
 ## [Upcoming]
 
+* Fix commit check logic when checking for non-Shepherd commits
+
 ## v1.0.0
 
-* Initial public releast to npm :tada:
\ No newline at end of file
+* Initial public releast to npm :tada:
```
borgstrom: nit: spelling

parshap:
I pushed another change to how the commit difference is calculated to use `HEAD..origin/branch` instead of `...` which is what simple-git does by default: https://github.com/steveukx/git-js/blob/144322bd03d9f38b7a5ed721a18b5eee862a7ba4/src/git.js#L1251

Issue #56, opened 2018-09-24
Tries to checkout the same repo multiple times


franjohn21:
If you're search query contains multiple matches for a given repo, that repo attempts to be checked out multiple times.

```
[NerdWallet/credit-cards-client] 27/278
✔ Checked out repo
> Running should_migrate steps
$ grep -rl -m 1 '.' -e "cdn.nerdwallet.com" --exclude-dir=node_modules | head -1
./client/components/pre-auth-modal/pre-auth-modal.less
Step "grep -rl -m 1 '.' -e "cdn.nerdwallet.com" --exclude-dir=node_modules | head -1" exited with 0
$ git log -1 --format=%cd | grep 2018
Sat Sep 22 04:46:19 2018 +0000
Step "git log -1 --format=%cd | grep 2018" exited with 0
✔ Completed all should_migrate steps successfully
> Running post_checkout steps
✔ Completed all post_checkout steps successfully

[NerdWallet/credit-cards-client] 28/278
✔ Checked out repo
> Running should_migrate steps
$ grep -rl -m 1 '.' -e "cdn.nerdwallet.com" --exclude-dir=node_modules | head -1
./client/components/pre-auth-modal/pre-auth-modal.less
Step "grep -rl -m 1 '.' -e "cdn.nerdwallet.com" --exclude-dir=node_modules | head -1" exited with 0
$ git log -1 --format=%cd | grep 2018
Sat Sep 22 04:46:19 2018 +0000
Step "git log -1 --format=%cd | grep 2018" exited with 0
✔ Completed all should_migrate steps successfully
> Running post_checkout steps
✔ Completed all post_checkout steps successfully
```

FoodProduct:
I think this is because shepherd is running [a code search](https://github.com/NerdWalletOSS/shepherd/blob/master/src/adapters/github.ts#L54) for github. Should this be a repo search? i.e. `this.octokit.search.code` => `this.octokit.search.repos`?

Issue #57, opened 2018-09-24
Rate limiting issues pulling from GitHub


FoodProduct:
When fetching a very long list of matching repos from GitHub, a consumer will run into a rate limiting error during `shepherd checkout`.
```
dbarnes:shepherd-migrations dbarnes$ shepherd checkout migrations/2018-09-cdn-url-updates/
{ HttpError: {
  "documentation_url": "https://developer.github.com/v3/#abuse-rate-limits",
  "message": "You have triggered an abuse detection mechanism. Please wait a few minutes before you try again."
}

    at response.text.then.message (/srv/nerdwallet/shepherd/node_modules/@octokit/rest/lib/request/request.js:72:19)
    at process._tickCallback (internal/process/next_tick.js:103:7)
  name: 'HttpError',
  code: 403,
  status: undefined,
  headers:
   { 'access-control-allow-origin': '*',
     'access-control-expose-headers': 'ETag, Link, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval',
     connection: 'close',
     'content-encoding': 'gzip',
     'content-security-policy': 'default-src \'none\'; base-uri \'self\'; block-all-mixed-content; connect-src \'self\' uploads.github.com status.github.com collector.githubapp.com api.github.com www.google-analytics.com github-cloud.s3.amazonaws.com github-production-repository-file-5c1aeb.s3.amazonaws.com github-production-upload-manifest-file-7fdce7.s3.amazonaws.com github-production-user-asset-6210df.s3.amazonaws.com; font-src assets-cdn.github.com; form-action \'self\' github.com gist.github.com; frame-ancestors \'none\'; frame-src render.githubusercontent.com; img-src \'self\' data: assets-cdn.github.com media.githubusercontent.com camo.githubusercontent.com identicons.github.com collector.githubapp.com avatars0.githubusercontent.com avatars1.githubusercontent.com avatars2.githubusercontent.com avatars3.githubusercontent.com github-cloud.s3.amazonaws.com; manifest-src \'self\'; media-src \'none\'; script-src assets-cdn.github.com; style-src \'unsafe-inline\' assets-cdn.github.com',
     'content-type': 'application/json; charset=utf-8',
     date: 'Mon, 24 Sep 2018 21:30:00 GMT',
[...]

PR #59, opened 2018-09-24
adding github rate failure support by abiding by retry-after

FoodProduct:
Fix for https://github.com/NerdWalletOSS/shepherd/issues/57
Also should fix https://github.com/NerdWalletOSS/shepherd/issues/56

### DIFF
--- a/src/util/octokit.ts
+++ b/src/util/octokit.ts
@@ -27,6 +27,8 @@ export const paginate = (
         console.info(`\nHit rate limit; waiting ${retryAfter} seconds and retrying.`);
         await wait(retryAfter * 1000);
         continue;
+      } else {
+        throw e;
       }
     }
   }
--- a/src/util/octokit.ts
+++ b/src/util/octokit.ts
@@ -1,8 +1,8 @@
 import Octokit from '@octokit/rest';
+import { RetryMethod } from '../adapters/base';
 
 type DataExtractor = (d: any) => any[];
 type Method = (opts: any) => Promise<any>;
-type RetryMethod = (num: number) => any;
 
 const wait = (timeout: number) => new Promise((resolve) => setTimeout(resolve, timeout));
 
--- a/src/util/octokit.ts
+++ b/src/util/octokit.ts
@@ -1,8 +1,8 @@
 import Octokit from '@octokit/rest';
-import { ILogger as logger } from '../logger';
 
 type DataExtractor = (d: any) => any[];
 type Method = (opts: any) => Promise<any>;
+type RetryMethod = (num: number) => any;
 
 const wait = (timeout: number) => new Promise((resolve) => setTimeout(resolve, timeout));
 
@@ -10,6 +10,7 @@ export const paginate = (
   octokit: Octokit,
   method: Method,
   extractItems: DataExtractor = (d: any) => d,
+  onRetry: RetryMethod,
 ) => async (options: any): Promise<any[]> => {
   let response: any = await method({
     ...options,
@@ -28,7 +29,7 @@ export const paginate = (
         if (Number.isNaN(retryAfter)) {
           throw e;
         }
-        logger.info(`\nHit rate limit; waiting ${retryAfter} seconds and retrying.`);
+        onRetry(retryAfter);
         await wait(retryAfter * 1000);
         continue;
       } else {
@@ -42,4 +43,5 @@ export const paginate = (
 // Search responses have a slightly different structure than normal ones, so we
 // need to extract the items from a different key
 const extractSearch = (data: any) => data.items;
-export const paginateSearch = (octokit: Octokit, method: Method) => paginate(octokit, method, extractSearch);
+export const paginateSearch = (octokit: Octokit, method: Method, onRetry: Method) =>
+    paginate(octokit, method, extractSearch, onRetry);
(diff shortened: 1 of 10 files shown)


src/util/octokit.ts
```diff
@@ -18,9 +18,19 @@ export const paginate = (
...
-
-    response = await octokit.getNextPage(response); // eslint-disable-line no-await-in-loop
-    data = data.concat(extractItems(response.data));
+    try {
+      response = await octokit.getNextPage(response); // eslint-disable-line no-await-in-loop
+      data = data.concat(extractItems(response.data));
+    } catch (e) {
+      if (e.headers && e.headers['retry-after']) {
+        const retryAfter = e.headers['retry-after'];
+        console.info(`\nHit rate limit; waiting ${retryAfter} seconds and retrying.`);
```
FoodProduct: Linter is erroring on `console.info()` here but I want to give the user feedback when this happens (since it likely means +60s each time). What is the recommended approach?

src/util/octokit.ts
```diff
@@ -18,9 +18,19 @@ export const paginate = (
   while (octokit.hasNextPage(response)) {
     // Avoid GitHub's "abuse detection mechanisms"
     await wait(2000);
```
FoodProduct: Should this wait be left at 2s or should this start being more aggressive now that the rate limiting catch is in place?

src/util/octokit.ts
parshap: We'll want to use the `logger` object like we do here: https://github.com/NerdWalletOSS/shepherd/blob/a5a092f2342f40dc20272eeee174a01d3332c3cb/src/commands/checkout.ts#L52

We don't have access to that in this helper. I think the best way to do this is to pass a `onRetry` callback from `commands/checkout.ts` through to `adapters/github.ts` all the way down to this helper so we can notify the cli command layer (which is responsible for input/output) about when a retry is happening.

src/util/octokit.ts
```diff
@@ -18,9 +18,19 @@ export const paginate = (
...
-
-    response = await octokit.getNextPage(response); // eslint-disable-line no-await-in-loop
-    data = data.concat(extractItems(response.data));
+    try {
+      response = await octokit.getNextPage(response); // eslint-disable-line no-await-in-loop
+      data = data.concat(extractItems(response.data));
+    } catch (e) {
+      if (e.headers && e.headers['retry-after']) {
+        const retryAfter = e.headers['retry-after'];
```
parshap: We should parse this to a number and handle invalid values. I suggest `Number(e.headers['retry-after']))` to parse the number and throwing if `Number.isNaN(retryAfter)` to handle invalid values.

src/util/octokit.ts
FoodProduct: Instead of passing an `onRetry`, I imported `logger` directly. Thoughts?

FoodProduct: Orrr typescript doesn't like that for some reason.

parshap:
I have no idea why you're getting the CI error, seems unrelated to the change here:

```
    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/adapters/github.ts:193:55 - error TS2339: Property 'mergeable_state' does not exist on type 'PullRequestsGetResponse'.
    193       } else if (pullRequest.mergeable && pullRequest.mergeable_state === 'clean') {
                                                              ~~~~~~~~~~~~~~~
```

src/util/octokit.ts
parshap: I'm not completely opposed, but the thought was to keep the adapters and utilities clean of logging so they could potentially be used in other contexts than the CLI.

src/util/octokit.ts
```diff
@@ -1,15 +1,16 @@
 import Octokit from '@octokit/rest';
-import { ILogger as logger } from '../logger';
 
 type DataExtractor = (d: any) => any[];
 type Method = (opts: any) => Promise<any>;
+type RetryMethod = (num: number) => any;
```
parshap: Should this come from `base.ts`?

parshap: Should this come from `base.ts`?

parshap:
Looks great, we just need to figure out the TS error now.

FoodProduct:
TS is bad plz halp @parshap

parshap:
Did some digging with Martin's help and concluded the following:

* @octokit/rest generates its type definitions via build-time script.
* At some point (in a minor version) they went from using `any` to specific types for API responses, which now does *not* include `mergeable_state` in the type definition.
* Build fails in node 6 only because it uses npm 3 without package-lock and thus gets the new version of @octokit/rest with type changes.
* We could use a hacky script to edit the type definition file at build time using `sed` or something, but that's probably too hacky.
* We might be able to "amend" the type definition using `extends`, I'll look into TS docs for this.
* Worst case, we should be able to use a `// @ts-disable` comment to disable type checking on that line.

parshap:
Based on some [light research](https://stackoverflow.com/questions/23217334/how-do-i-extend-a-typescript-class-definition-in-a-separate-definition-file) it seems you can't extend an existing `type` definition to amend a new property. You can do this with `interface`, but the octokit/rest module uses `type` here.

I'm going to use a `// @ts-ignore` comment and open an issue in octokit/rest.

parshap:
https://github.com/octokit/rest.js/issues/1038

PR #63, opened 2018-10-05
fix: spelling in list command

piperchester:


franjohn21:
We should have a CI job that spellchecks stuff like this somehow

PR #75, opened 2019-02-28
GitHub adapter: add ability to checkout all of an organizations repos

benjroy:
Github code search can be unreliable, returning an indeterminate list of results for the same `search_query` each time it is run.

In cases like that, it might be preferable to take a more naive approach:  List every repo in my organization.  It will be slower (must checkout all repos before running checks to see if they match the target for the migration being run), but it will get rid of question of whether or not you truly matched every repo intended.

Version `1.2.0`


src/adapters/github.ts
```diff
@@ -50,12 +50,55 @@ class GithubAdapter extends GitAdapter {
...
+  //   // list all of an orgs repos
+  //   if (org) {
+  //     if (search_query) {
+  //       throw new Error('Cannot use both "org" and "search_query" in github adapter. Pick one.');
+  //     }
+  //      return this.getOrgRepos(org, onRetry);
+  //   }
+  //   // otherwise, github code search query.  results are less reliable
+  //   return this.getCodeSearchRepos(search_query, onRetry);
+  // }
```
parshap: Remove?

docs/tutorial.md
```diff
@@ -85,6 +85,16 @@ adapter:
   search_query: repo:YOURUSERNAME/shepherd-demo path:/ filename:.eslintrc
 ```
 
+**Note:** you can provide an `org` instead of `search_query`. This will return a list of every visible repo in a github organization.
```
parshap: 👍 

Might be worth mentioning in the readme as well next to where `search_query` is currently mentioned.

src/adapters/github.ts
```diff
@@ -50,12 +50,55 @@ class GithubAdapter extends GitAdapter {
...
-      q: this.migrationContext.migration.spec.adapter.search_query,
-    });
-    const repoNames = searchResults.map((r: any) => r.repository.full_name).sort();
+    const { org, search_query } = this.migrationContext.migration.spec.adapter;
+    let repoNames = [];
+
+    // list all of an orgs repos
+    if (org) {
+      if (search_query) {
+        throw new Error('Cannot use both "org" and "search_query" in github adapter. Pick one.');
```
parshap: Nice error handling.

docs/tutorial.md
parshap: Might also be worth adding some context around _why_ you'd use `org` instead of `search_query` (if you're getting inconsistent search results or have more complicated filtering logic you'd like to run solely in `should_migrate`).

src/adapters/github.ts
benjroy: oh.  wow.  yes.  thank you

docs/tutorial.md
piperchester:
```suggestion
**Note:** you can provide an `org` instead of `search_query`. This will return a list of every visible repo in a GitHub organization.
```

src/adapters/github.ts
piperchester:
```suggestion
        throw new Error('Cannot use both "org" and "search_query" in GitHub adapter. Pick one.');
```

src/adapters/github.ts
```diff
@@ -50,12 +50,55 @@ class GithubAdapter extends GitAdapter {
...
+      }
+      const repos = await paginate(this.octokit, this.octokit.repos.listForOrg, undefined, onRetry)({
```
piperchester: Nit: I think this is outside the scope of this PR but just wanted pose the question 💭 

Perhaps in the future we could refactor [the signature of `paginate`](https://github.com/NerdWalletOSS/shepherd/blob/master/src/util/octokit.ts#L12) so that at the call site we could omit `extractorItems`. And it'd have more symmetry with `paginateSearch`.

nwalters512: +1 - keeping optional parameters at the end of the parameter list seems sensible!

benjroy: +1 - could also go for an interface that takes a single argument object

That's for a different PR for a different time 👍

nwalters512:
> guthub-adapter-org-repos-list

Loving the branch name 😉

PR #88, opened 2019-12-08
Enable adapters to add environment variables

TheSavior:
I'd like my GitHub PR messages to include permalinks, so they are rendered like this:

https://github.com/NerdWalletOSS/shepherd/blob/8048f846335ab5b16c6931b01c8974d2a8914406/package.json#L2-L8

In order to do this my PR script needs to be able to generate URLs that look like:
```
https://github.com/NerdWalletOSS/shepherd/blob/8048f846335ab5b16c6931b01c8974d2a8914406/package.json#L2-L8
```

This requires the message generator to have access to the owner, repo, and revision being operated on. As far as I can tell that isn't currently possible. 

It looks like there is a precedent for some of this information as environment variables at script time, including the `branch` of the current repo:
https://github.com/NerdWalletOSS/shepherd/blob/8048f846335ab5b16c6931b01c8974d2a8914406/src/util/exec-in-repo.ts#L18-L24

Instead of hard coding more things in here that are specific to GitHub or Git, I modified the adapters to be able to specify additional environment variables. Perhaps the base branch environment variable should get moved into the `git` adapter with the `revision` env var added in this PR.

### DIFF
--- a/src/util/exec-in-repo.ts
+++ b/src/util/exec-in-repo.ts
@@ -8,11 +8,18 @@ interface IExecRepoResult {
   childProcess: ChildProcess;
 }
 
-export default (context: IMigrationContext, repo: IRepo, command: string): IExecRepoResult => {
+export default async (
+  context: IMigrationContext,
+  repo: IRepo,
+  command: string,
+): Promise<IExecRepoResult> => {
   const repoDir = context.adapter.getRepoDir(repo);
   const dataDir = context.adapter.getDataDir(repo);
   const baseBranch = context.adapter.getBaseBranch(repo);
   const migrationDir = context.migration.migrationDirectory;
+
+  const adapterEnvironmentVars = await context.adapter.getEnvironmentVariables(repo);
+
   const execOptions = {
     cwd: repoDir,
     env: {
@@ -21,6 +28,7 @@ export default (context: IMigrationContext, repo: IRepo, command: string): IExec
       SHEPHERD_DATA_DIR: dataDir,
       SHEPHERD_MIGRATION_DIR: migrationDir,
       SHEPHERD_BASE_BRANCH: baseBranch,
+      ...adapterEnvironmentVars,
     },
     shell: true,
     capture: [ 'stdout', 'stderr' ],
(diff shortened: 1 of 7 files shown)


src/util/exec-in-repo.ts
```diff
@@ -8,11 +8,17 @@ interface IExecRepoResult {
   childProcess: ChildProcess;
 }
 
-export default (context: IMigrationContext, repo: IRepo, command: string): IExecRepoResult => {
+export default async (
```
TheSavior: This was previously sync, requiring that the potentially long running `spawn` promise was returned to be able to be awaited at a later point. 

I kept that behavior the same, but need to get the environment variables first. So this method essentially now lets the implementation await some things, but not the spawn. Neat.

## My other approach
I initially tried to keep this method sync and put the async `getEnvrionmentVariables` call inside the promise by doing something like this:

```
const promise = adapterEnvironmentVarsPromise.then(
    (adapterEnvironmentVars: IEnvironmentVariables) => ({
        cwd: repoDir,
        env: {
          ...process.env,
          SHEPHERD_REPO_DIR: repoDir,
          SHEPHERD_DATA_DIR: dataDir,
          SHEPHERD_MIGRATION_DIR: migrationDir,
          SHEPHERD_BASE_BRANCH: baseBranch,
          ...adapterEnvironmentVars,
        },
        shell: true,
        capture: ['stdout', 'stderr'],
      })
  ).then((execOptions: import("child_process").SpawnOptions) => {
    return spawn(command, [], execOptions)
  }) as ChildProcessPromise;
```

However, the return of this function is 
```
return {
  promise,
  childProcess: promise.childProcess,
};
```
and the spawn promise won't have been created at the time this function returns, so `childProcess` doesn't exist on the `promise` variable.

Alas, I ended up to where I am in this PR.

*Note: there is only one callsite of this method, so there doesn't seem to be much risk of converting this to be async*

parshap: The interface is a little awkward, but I think what you have is fine.

TheSavior:
I've updated the documentation

PR #100, opened 2020-02-18
Remove deprecated Github auth mechanisms

### LINKED ISSUE #99: Using deprecated auth method
Hi!
It seems shepherd is using the auth method Github is in process of deprecating: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param (I get an email from Github about using it when running shepherd).
Is there some dependency you need to bump, or is this in shepherd itself?

carlpett:
Fixes #99 

I originally just wanted to fix the token auth [1], but as I understand [2], the existing netrc fallback to user/password is **also** being deprecated... So I removed that entirely. This means that this PR is a breaking change though! :rotating_light: 
Warning for me having quite few clues when it comes to Typescript, and no working dev environment either. So this is coded in the dark, and probably needs some fixup. Sorry about that!

ping @nwalters512 

- [1] https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param
- [2] https://developer.github.com/changes/2020-02-14-deprecating-password-auth/


nwalters512:
Thanks for the PR! The `auth` API wasn't added until `@octokit/rest` v16.13.0 - you'll have to bump Octokit and check for any breaking changes. That should fix the TS error that's showing up in CI.

Regarding the removing of `.netrc` support, NerdWallet uses `.netrc` internally to share GitHub auth token between tools, so keeping this around is still important for us. It should be pretty simple to keep support for this around; it should just involve something like this:

```ts
this.octokit = new Octokit({
  auth: {
    username: '...',
    password: '...',
  },
});
```

Once support for basic auth is officially removed, we can look at officially removing this feature.

carlpett:
Hey,
Sorry, I think I'll just have to step away from this one :disappointed: Setting up a working nodejs environment will take more time than I have available right now. Hope someone else has a few spare cycles!

Issue #108, opened 2020-04-08
Dependency Management Part II: Update jest


aorinevo:
## Introduction

This issue picks up where issue #103 leaves off.  In particular, this issue addresses bringing jest and jest related dependencies up-to-date.

## Proposed Change
- Remove `jest-plugin-fs` dependency in favor of using local artifacts
  - Idea here is to use a few publicly accessible github repos
- Test each shepherd cli command

## ROI
- Resolves remaining 206 vulnerabilities
- Unit tests more aligned with user actions

aorinevo:
PR #113 is ready for review.

aorinevo:
Screenshots:
<img width="682" alt="Screen Shot 2020-04-09 at 3 06 18 AM" src="https://user-images.githubusercontent.com/3240137/78867430-172acb00-7a0f-11ea-8af0-c1aa3439a5b2.png">
<img width="900" alt="Screen Shot 2020-04-09 at 3 06 40 AM" src="https://user-images.githubusercontent.com/3240137/78867460-2447ba00-7a0f-11ea-9935-1c0f3f43e77b.png">

aorinevo:
TravisCI is failing.  I’ll take a look at it tomorrow.

aorinevo:
Correpsonding PR https://github.com/NerdWalletOSS/shepherd/pull/113

PR #113, opened 2020-04-09
chore: update jest

aorinevo:
## Changes
- closes https://github.com/NerdWalletOSS/shepherd/issues/108
- update jest to v25
- remove `jest-plugin-fs` dependency

## Notes
- with this PR, vulnerabilities go down to 0


src/util/persisted-data.ts
```diff
@@ -15,7 +15,7 @@ const jsonStringify = (data: any) => JSON.stringify(data, undefined, 2);
  */
 const migrateToJsonIfNeeded = async (migrationContext: IMigrationContext) => {
   const legacyFile = getLegacyRepoListFile(migrationContext);
-  if (await fs.existsAsync(legacyFile)) {
```
aorinevo, line 18: existsAsync is deprecated for node 8+

src/cli.ts
```diff
@@ -24,7 +24,9 @@ import reset from './commands/reset';
 
 import ConsoleLogger from './logger';
 
-const shepherdDir = path.join(homedir(), '.shepherd');
+const { DOT_SHEPHERD_DIRECTORY_LOCATION } = process.env;
```
aorinevo, line 27: TravisCI needs to run with env var set.

src/util/persisted-data.ts
nwalters512: This is provided by `fs-extra-promise`, it isn't the version from the native `fs` module. I'd prefer to keep using async fs methods everywhere if possible!

.travis.yml
```diff
@@ -3,15 +3,15 @@ node_js:
   - 8
   - 10
   - 12
-cache:
-  directories:
-    - node_modules
```
nwalters512, line 8: Can you explain why this was removed?

package.json
```diff
@@ -43,7 +43,8 @@
     }
   },
   "files": [
-    "lib"
+    "lib",
+    "src"
```
nwalters512, line 47: Is there a reason that `src` should be distributed with the package?

package.json
```diff
@@ -63,19 +64,17 @@
     "simple-git": "^1.132.0"
   },
   "devDependencies": {
-    "@types/fs-extra": "^5.0.4",
-    "@types/fs-extra-promise": "^1.0.7",
+    "@types/fs-extra": "^8.1.0",
```
nwalters512, line 67: If you're bumping types, can you keep the dependencies themselves in sync? E.g. I would expect `fs-extra` to have the same semver range as its type definitions.

src/cli.ts
nwalters512: Nice! Nit: consider just `DOT_SHEPHERD_DIRECTORY` - `LOCATION` feels redundant to me.

src/tests/artifacts/.shepherd/adds-checked-out-repo/repos-org.json
```diff
@@ -0,0 +1 @@
+[{"name":"test","owner":"NerdWallet"}]
```
nwalters512, line 1: Nit: could you put these files in a `fixtures` dir in the root of the repo instead of in `src/tests/artifacts`?

src/tests/artifacts/.shepherd/prefs.yaml
```diff
@@ -0,0 +1 @@
+workingDirectory: /Users/aorinevo/Repositories/nerdwallet/shepherd/src/tests/artifacts/.shepherd
```
nwalters512, line 1: Nit: this path looks specific to your machine - consider adding this file to `.gitignore`.

package.json
aorinevo: Will do.

src/tests/artifacts/.shepherd/prefs.yaml
aorinevo: Will do.

aorinevo: I think I'll just remove it.

src/tests/artifacts/.shepherd/adds-checked-out-repo/repos-org.json
aorinevo: There will be a trade-off here.  Since the test files require source files, typescript will throw an error TS6059 (i.e. File '...' is not under 'rootDir' ''. 'rootDir' is expected to contain all source files). To overcome this, we can remove 'rootDir' from tsconfig options; however, that would mean setting `bin.shepherd` in package.json equal to './lib/src/cli.js'.

Thoughts?

package.json
aorinevo: I believe this can be removed and an artifact of trying to add version command which is or will be addressed in a separate PR.

.travis.yml
aorinevo: Caching node_modules directory had unexpected side effects on CI.  In particular, I got the following error:
```bash
The command "eval npm ci  " failed. Retrying, 2 of 3.
```

Removed cache as per: https://github.com/travis-ci/travis-ci/issues/10113#issuecomment-421121316

src/util/persisted-data.ts
aorinevo: `fs-extra-promise` has fallen out of maintenance (see https://github.com/overlookmotel/fs-extra-promise#readme).  I'd prefer to find an alternative to `fs-extra-promise`.  

Perhaps use `fs.access` (see https://nodejs.org/api/fs.html#fs_fspromises_access_path_mode)?

aorinevo: Node 8+ supports `fs.access`.

aorinevo:
> Gave this a first pass! Tests are failing on my machine, looks like they are on CI as well.
> 
> I'd appreciate a little more context on these changes - what's the goal with moving from a virtual filesystem with `jest-plugin-fs` to working with actual files on disk? Can you explain the `src/tests/` directory a bit?

There are two main drivers for removing `jest-plugin-fs`:
- `jest-plugin-fs` has fallen out of maintenance
- reduces the overhead in writing tests for the cli commands themselves, similar to what was implemented in Handlebars (see https://github.com/wycats/handlebars.js/tree/master/spec).

If we are steadfast on using mocks, maybe we can switch to using jest's mocks directly? (see https://jestjs.io/docs/en/manual-mocks)

nwalters512:
> If we are steadfast on using mocks

Definitely not the case! I was just curious for the rationale behind moving away from them. I think the potential to have CLI integration tests is very intriguing; happy to move to working with a real filesystem to enable that.

src/tests/artifacts/.shepherd/adds-checked-out-repo/repos-org.json
nwalters512: Could we just do a `fs.readFile` + `JSON.parse` instead of a `require`? That should let us avoid any TypeScript problems.

.travis.yml
nwalters512: Gotcha. That's unfortunate! Thanks for fixing 🙏

src/util/persisted-data.ts
nwalters512: Looks like `fs-extra` actually has native promise support! We should definitely move to that. If you wouldn't mind, could you back this change out of this PR so we can move the whole project to `fs-extra` in a separate PR?

aorinevo: Will do :)

src/tests/artifacts/.shepherd/adds-checked-out-repo/repos-org.json
aorinevo: Absolutely!  Will do.

aorinevo:
@parshap and @nwalters512, this PR is ready for re-review.

src/util/persisted-data.ts
```diff
@@ -15,10 +15,10 @@ const jsonStringify = (data: any) => JSON.stringify(data, undefined, 2);
...
-    const data = yaml.safeLoad(await fs.readFileAsync(legacyFile, 'utf8'));
-    await fs.writeFileAsync(getRepoListFile(migrationContext), jsonStringify(data));
-    await fs.unlinkAsync(legacyFile);
+  if (await fs.pathExists(legacyFile)) {
+    const data = yaml.safeLoad(await fs.readFile(legacyFile, 'utf8'));
+    await fs.outputFile(getRepoListFile(migrationContext), jsonStringify(data));
```
nwalters512, line 20: Nice, TIL about `outputFile`

src/util/persisted-data.test.ts
```diff
@@ -1,94 +1,106 @@
...
+    const workingDirectory = path.join(__dirname, '../../fixtures/artifacts/.shepherd/no-repos-json/');
+    const repos = await loadRepoList(makeContext(workingDirectory));
     expect(repos).toEqual(null);
   });
 
   it('migrates from a YAML file to a JSON file', async () => {
-    const repos = await loadRepoList(makeContext());
-    expect(repos).toEqual([{ owner: 'NerdWallet', name: 'test' }]);
+    const workingDirectory = path.join(__dirname, '../../fixtures/artifacts/.shepherd/yaml-to-json/');
+    const source = path.join(workingDirectory,'repos-artifact.yml');
```
nwalters512, line 38: Formatting nit: can you ensure there are spaces between function arguments?

We should add Prettier to this repo in another PR to avoid formatting comments like this 🙂

src/util/persisted-data.test.ts
```diff
@@ -1,94 +1,106 @@
...
-    '/migration/repos.json': JSON.stringify(reposFixture),
-  }));
-  afterEach(() => jestFs.restore());
 
   it('loads repo list from a file', async () => {
-    const repos = await loadRepoList(makeContext());
-    expect(repos).toEqual([{ owner: 'NerdWallet', name: 'test' }]);
+    const workingDirectory = path.join(__dirname, '../../fixtures/artifacts/.shepherd/load-repo-from-file/');
+    const repos = await loadRepoList(makeContext(workingDirectory));
+    // fs.unlinke(path.join(workingDirectory,'/repos.json'));
```
nwalters512, line 26: Can you remove this line if it's unnecessary?

src/util/persisted-data.test.ts
aorinevo: Absolutely!  I was thinking of doing that along with yamllinter.

src/util/persisted-data.test.ts
aorinevo: Will do.

src/util/persisted-data.test.ts
nwalters512: Yes! Linting as a whole could use some love. ESLint is now the preferred way to lint TypeScript, so we should move to that from TSLint.

aorinevo: Yay! ESlint ❤️

aorinevo:
🚢

Issue #145, opened 2020-04-18
Feature request - make "branding" optional


jugaltheshah:
I wanted to open a discussion on the possibility of making the Shepherd "branding" during PR creation optional via env var / CLI option. The commit title in particular breaks things looking for semantic commit messages.

aorinevo:
@jugaltheshah, I'm definitely on board with this.  I'd be happy to review a PR for this, if you'd like to take this on.

Another thing to consider and potential for future work.  I think we'll need to implement configuration management that goes beyond `env vars`.  That is, allow users to configure the behavior of Shepherd via some config file, say `.shepherdrc.json` with prioritization given to values set in `.shepherdrc.json` over env vars.

ROI:
- Persistence of configurations across terminal sessions
- Retradable as a way to feature flag
- Componentizes configuration management

(I'll open that a separate issue)

parshap:
I’m on board with that too. It might be good to just remove the prefix in Shepherd and rely on `shepherd.yml` specifying the title in full.

Agree with a better config system too. I wrote some thoughts on this here: https://github.com/NerdWalletOSS/shepherd/issues/91

My vision is a unified “cascading” config system, where every option can either be set at a global level (e.g., `~/shepherd.yml`), a migration file, or as command line options (with precedence working in that order too). I think a bifurcated system where there’s a separate set of “config” from migration options you set in shepherd.yml might make sense too, but I think it will be hard to predict what people will want to control globally vs on a per-migration basis.

PR #228, opened 2020-08-13
fix: suffix [shepherd] to not break semantic commit

DiegoRBaquero:


aorinevo:
Taking a look.

aorinevo:
Yea, looks good to me.  

This is loosely related to the branding open issue, in that it touches the same line of code (see https://github.com/NerdWalletOSS/shepherd/issues/145).

nwalters512:
> This is loosely related to the branding open issue, in that it touches the same line of code

Good point - I'm honestly in favor of just dropping the "[shepherd]" label altogether - I think the footer Shepherd adds to PR messages is enough in the way of branding + identifying the PR as from Shepherd.

aorinevo:
On second hand, can't we just drop `[shepherd]`?  Users can add the prefix to via `spec.title`.

src/adapters/git.ts
```diff
@@ -61,7 +61,7 @@ abstract class GitAdapter implements IRepoAdapter {
   public async commitRepo(repo: IRepo): Promise<void> {
     const { migration: { spec } } = this.migrationContext;
     await this.git(repo).add('.');
-    await this.git(repo).commit(`[shepherd] ${spec.title}`);
+    await this.git(repo).commit(`${spec.title} [shepherd]`);
```
parshap, line 64: Any opposition to removing the branding entirely?

```suggestion
    await this.git(repo).commit(spec.title);
```

At NerdWallet we still like to call out shepherd migrations in the title, but we can prefix `[shepherd]` manually in our spec titles.

nwalters512: I think we've got consensus, let's do it!

aorinevo:
See https://github.com/NerdWalletOSS/shepherd/blob/a849b16616a56dfc0a25907810899cc37db09bae/src/adapters/github.ts#L337

parshap:
Good catch @aorinevo, we do that to determine whether it's safe to force-push over existing commits on a branch (we do it only if they are all Shepherd commits). Sounds like we'll need to keep `[shepherd]` in the title somewhere and update that logic to check for `[shepherd]` anywhere instead of just at the beginning. If that makes sense to everyone, @DiegoRBaquero can you update the PR to include that?

DiegoRBaquero:
Updated :)

parshap:
Looks good. We can look at how we could potentially remove the branding entirely as a separate effort from this PR.

Will wait for any comments from @aorinevo or @nwalters512 before merging.

aorinevo:
That can be handled with the branding story by setting the default value to `[shepherd]`.  Consumer can override with empty string or their brand's name.

aorinevo:
Danke @DiegoRBaquero!

PR #324, opened 2020-12-04
add incomplete github api search detection

danpilch:
Shepherd currently does not provide a way to fail or warn that a github api search is incomplete. This PR provides a `warning` to the user if the search is incomplete and a flag `shepherd checkout --fail-on-incomplete-search migration` to fail the migration. 

This is the first time I have ventured into Typescript so please do let me know if there's a better way to work in this functionality.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -9,7 +9,7 @@ export interface IEnvironmentVariables {
 export type RetryMethod = (opts: number) => any;
 
 interface IRepoAdapter {
-  getCandidateRepos(onRetry: RetryMethod): Promise<IRepo[]>;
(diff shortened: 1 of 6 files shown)


aorinevo:
Hi @danpilch, apologies for the delayed review of this PR.  Very much appreciate the contribution.  

With respect to the PR, we would need unit tests for the changes.

With respect to the feature, it seems useful.  Mind opening an issue against this PR including the summary from this PR, a list of use cases, and a list of benefits?

@nwalters512, thoughts?

danpilch:
@aorinevo I have created issue https://github.com/NerdWalletOSS/shepherd/issues/401

aorinevo:
@danpilch, can you add a unit test for this changes and update the PR with latest from master?

Issue #354, opened 2021-01-05
Migrate @hapi/joi back to joi


aorinevo:
## Introduction

HapiJS has descoped `joi` from `@hapi` which means that we will stop receiving auto-updates for minor and patch versions.  In order to fix this issue, we should migrate back to installing directly from `joi`.

## References
1. https://github.com/NerdWalletOSS/shepherd/pull/353/files#r551693336

aorinevo:
Closed by #358

PR #358, opened 2021-01-09
chore: update @hapi/joi to joi

jugaltheshah:
Addresses #354

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,13 +46,13 @@
     "lib"
   ],
   "dependencies": {
-    "@hapi/joi": "^17.1.1",
     "@octokit/rest": "^15.18.1",
     "@types/js-yaml": "^3.12.6",
     "chalk": "^4.1.0",
     "child-process-promise": "^2.2.1",
     "commander": "^6.2.1",
     "fs-extra": "^9.0.1",
+    "joi": "^17.3.0",
     "js-yaml": "^3.14.1",
     "lodash": "^4.17.19",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -65,7 +65,6 @@
     "@semantic-release/changelog": "^5.0.1",
     "@semantic-release/git": "^9.0.0",
     "@types/fs-extra": "^9.0.6",
-    "@types/hapi__joi": "^17.1.6",
```
aorinevo, line 68: Good catch - Joi includes types within then package.

aorinevo (approved):

src/util/migration-spec.test.ts
```diff
@@ -96,7 +96,7 @@ describe('validateSpec', () => {
   });
 
   it('rejects a spec with a missing adapter type', () => {
-    const spec = cloneDeep(baseSpec);
+    const spec: any = cloneDeep(baseSpec);
```
jugaltheshah, line 99: This was needed as I was getting the following error running `tsc`:
```bash
src/util/migration-spec.test.ts:100:12 - error TS2790: The operand of a 'delete' operator must be optional.

100     delete spec.adapter.type;
               ~~~~~~~~~~~~~~~~~
```
Basically it's saying that `spec.adapter.type` was declared as required instead of optional, so it can't be deleted. We of course _can't_ make it optional, as that is the point of the test, hence this.

PR #360, opened 2021-01-09
feat: update octokit

jugaltheshah:
Deprecated - Github username & password auth:
As Github has [deprecated](https://docs.github.com/en/free-pro-team@latest/rest/overview/other-authentication-methods#via-username-and-password) REST API auth via username & password, this no longer works as an auth mechanism. After this PR, shepherd will still support auth via `.netrc`, but the `.netrc.password` field should supply a token instead of a Github password.

Updates:
- `"@octokit/rest": "^15.18.1"` -> `"@octokit/rest": "^18.0.12",`,
  - octokit methods, e.g. `octokit.pullRequests.update` ->  `octokit.pulls.update`


src/adapters/github.ts
```diff
@@ -1,13 +1,14 @@
 /* eslint-disable class-methods-use-this */
-import Octokit from '@octokit/rest';
+/* eslint import/no-unresolved: [1, { ignore: ['dist-types/generated/method-types$'] }] */
```
jugaltheshah, line 2: Was having issues with the `RestEndpointMethods` import after this, hence this line. Error:

```bash
    src/adapters/github.ts:3:37 - error TS2307: Cannot find module '@octokit/plugin-rest-endpoint-methods/dist-types/generated/method-types' or its corresponding type declarations.

    3 import { RestEndpointMethods } from '@octokit/plugin-rest-endpoint-methods/dist-types/generated/method-types';
``` 

Could use some help resolving this error.

aorinevo:
@nwalters512, I recommend bumping the major version here and provide support for the last 2 major versions.  Thoughts?

src/adapters/github.test.ts
```diff
@@ -12,7 +12,7 @@ const mockMigrationContext = () => ({
   },
 });
 
-const noOpOnRetry = () => { /* do-nothing onRetry fn for mocks */ }
+// const noOpOnRetry = () => { /* do-nothing onRetry fn for mocks */ }
```
aorinevo, line 15: If no longer needed, let's remove.

src/adapters/github.ts
aorinevo: Taking a look.

aorinevo:
@jugaltheshah, as this is a breaking change, let's bump the major version.

src/adapters/github.ts
```diff
@@ -90,11 +83,12 @@ class GithubAdapter extends GitAdapter {
...
 
-      const searchResults = await paginateSearch(this.octokit, searchMethod, onRetry)({
-        q: search_query,
-      });
-
-      repoNames = searchResults.map((r: any) => _.get(r, fullNamePath) ).sort();
+      const searchResults: string[] = await this.octokit.paginate(
+        searchMethod,
+        { q: search_query },
+        (r: any) => r.items
```
aorinevo, line 89:
```suggestion
        (r: any) => r.data
```

aorinevo: Are we missing a unit test for this?

src/adapters/github.ts
aorinevo: I think you want to `import { RestEndpointMethodTypes } from '@octokit/plugin-rest-endpoint-methods'` instead (see [1]).

References
1. https://github.com/octokit/plugin-rest-endpoint-methods.js/#typescript

jugaltheshah: That's what I had initially but it was throwing all kinds of type errors.

jugaltheshah:
> @nwalters512, I recommend bumping the major version here and providing support for the last 2 major versions. Thoughts?

That was the intent with the "breaking change" in the PR description though come to think of it, I'm not sure if that's a custom option that this repo doesn't have enabled. Regardless, I'll update the PR title, which should do the same.

**Please be sure to squash & merge this PR so semantic bot will cut a new major release**

src/adapters/github.ts
aorinevo: I see.  Looks like `RestEndpointMethodTypes['search']['repos']` is not compatible with `octokit.search.repo`.

src/adapters/github.ts
jugaltheshah: This is intentional. Per the [octokit pagination doc](https://octokit.github.io/rest.js/v18#pagination):

> octokit.paginate() accepts the same options as octokit.request(). You can optionally pass an additional function to map the results from each response. The map must return a new value, usually an array with mapped data.

> Note: the map function is called with the { data, headers, status, url } response object. The data property is guaranteed to be an array of the result items, even for list endpoints that respond with an object instead of an array, such as the search endpoints.

In this case, I'm passing in that optional function to map results which is called with `{ data, headers, status, url }`.

jugaltheshah: Oops, I thought I had `.data`. Yes it should be data, resolving.

src/adapters/github.ts
aorinevo: I tried to trackdown where the incompatibility stems from but wasn't successful.  Will try again later.

aorinevo:
> > @nwalters512, I recommend bumping the major version here and providing support for the last 2 major versions. Thoughts?
> 
> That was the intent with the "breaking change" in the PR description though come to think of it, I'm not sure if that's a custom option that this repo doesn't have enabled. Regardless, I'll update the PR title, which should do the same.
> 
> **Please be sure to squash & merge this PR so semantic bot will cut a new major release**

Need to confirm that this is the expected behavior.

src/adapters/github.ts
jugaltheshah: E.g.
image

If you know how to fix please let me know. Updating the code was not too bad, it's these TS issues that I banged my head against for _far_ too long :cry:.

aorinevo: I guess in the interim, you could do the following:

```suggestion
      let fullNamePath: string;
      let searchResults: string[];

      switch (search_type) {
        case 'repositories':
          searchResults = await this.octokit.paginate(
            this.octokit.search.repos,
            { q: search_query },
            (r: any) => {
              return r.data
            }
          )
          fullNamePath = 'full_name'
          break;
        case 'code':
        default:
          // github code search query. results are less reliable
          searchResults = await this.octokit.paginate(
            this.octokit.search.code,
            { q: search_query },
            (r: any) => {
              return r.data
            }
          )
          fullNamePath = 'repository.full_name'
      }
```

aorinevo:
@jugaltheshah, mind squashing the commits down prior to merging this down.  In addition, can you follow the message commit format is specified in [1].

References
1. https://www.conventionalcommits.org/en/v1.0.0/

nwalters512:
> I recommend bumping the major version here and providing support for the last 2 major versions. Thoughts?

Is the breaking change the difference in how `.netrc` works? I'm on board with calling this v2, yeah.

By "providing support for the last 2 major versions", do you mean backporting future changes to v1? I don't think that'll be necessary.

jugaltheshah:
> > I recommend bumping the major version here and providing support for the last 2 major versions. Thoughts?
> 
> Is the breaking change the difference in how `.netrc` works? I'm on board with calling this v2, yeah.
> 
> By "providing support for the last 2 major versions", do you mean backporting future changes to v1? I don't think that'll be necessary.

That's an interesting question actually. I'm actually now questioning if we actually need to mark this as breaking. 

Yes, the reason this PR is currently breaking is the difference in how `.netrc` works. However, the [docs](https://docs.github.com/en/free-pro-team@latest/rest/overview/other-authentication-methods#via-oauth-and-personal-access-tokens) say you can use a token in place of a password for clients that only support basic auth. So I'm wondering, can we can keep the existing `.netrc` un/pw flow, but assume the password supplied is actually a token?

So then if we keep `.netrc` and use the supplied username & password (token), would this still be breaking (keeping in mind that using password as password hass _been_ broken)?

cc @aorinevo

src/adapters/github.ts
jugaltheshah: Haha I see your point about this maybe not being a fix :smile:. I think per our conversation, I'll try to go the `any` route first as this feels a bit duplicative for no real gain. Hopefully I don't fall down another TS rabbit hole. Thanks for the deep dive into types!

jugaltheshah: The `any` approach worked, resolving this.

aorinevo:
Netrc should still work, right?  It would potentially start to degrade during the brownout period and then be fully non-functional when username/password is deprecated.

I'm still in favor of doing a major version bump.

aorinevo:
> > I recommend bumping the major version here and providing support for the last 2 major versions. Thoughts?
> 
> Is the breaking change the difference in how `.netrc` works? I'm on board with calling this v2, yeah.
> 
> By "providing support for the last 2 major versions", do you mean backporting future changes to v1? I don't think that'll be necessary.

Security and bug fixes only for the last major version - no backporting of features/enhancements.

jugaltheshah:
> Netrc should still work, right? It would potentially start to degrade during the brownout period and then be fully non-functional when username/password is deprecated.
> 
> I'm still in favor of doing a major version bump.

Using username & password against the Github REST API was already deprecated as of November 13th, 2020 per [this doc](https://docs.github.com/en/free-pro-team@latest/rest/overview/other-authentication-methods#via-username-and-password). 

I think what's causing some confusion here is that Github deprecated using username:password auth, but _not_ basic auth via username:token with token used like a password. So if your Github username & password were user1:password123, using those against the Github API won't work. However, if you generate a personal access token on Gihub.com and use user1:token123 against the Github API, I believe this will still work (this is what I still need to confirm). They did this for compatibility with clients that only support basic auth. Of course _we_ don't actually need a username since tokens are associated with users already, so really we'd just need to assume a password provided is a token.

So we have a couple options:
1. Keep using `.netrc` but re-purpose the `login` field to be used as a token --> breaking, but I believe this is what [octokit/netrc](https://github.com/travi/octokit-auth-netrc) does. This is what this PR currently does also.
2. Keep using `.netrc`, keep reading from `password`, but assume password provided = Github token, _not_ Github password --> _not_ necessarily breaking*

* Not necessarily breaking from our perspective because as previously noted, this would have _already_ been broken as of mid-Nov of last year. If users currently have their Github password in `.netrc`, it shouldn't work, but if they update it to a personal access token and we go with option 2, it should work.

We could err on the side of caution here and release a major version regardless, but just wanted to present the options as I saw them.

aorinevo:
Ah, I see now.  That makes sense. 

I'm in favor of option 2 and keeping this a minor bump.

aorinevo (approved):

aorinevo:
Many thanks for this contribution @jugaltheshah! 🚢

Issue #361, opened 2021-01-10
Idea for improvement - use GraphQL


jugaltheshah:
I noticed that `shepherd` is doing a bunch of Github calls and getting a bunch of data that's mostly just discarded. Wanted to float the idea of switching to the Github GraphlQL API (via [octokit/graphql](https://github.com/octokit/graphql.js/#variables)) for at least some if not all calls, to optimize some of this.

For example:
- getPullRequestStatus - makes 2 Github calls - 1 call to get PR + 1 call to get PR status. I think using GraphQL we can get this down to 1 call, which would shave off ~50% of network time for that command, _for each repo_.
- getCandidateRepos (org query) - makes 1 call but the data returned is fairly large. I tried it on my own org and got back 423 repos at 2.1MB. I then filtered that data for just `full_name` and `archived` fields and the resulting file size is 24KB. So we're retrieving almost 90x the data actually needed and just discarding the rest! This is the case for a lot of the Github calls.

Wanted to get an idea of the level of interest before I make a PR, bc it's a decent amount of work. Here's the plan I had in mind.

Phase 1: split out all octokit calls into a new Github service
Phase 2: identify calls with most room for improvement and switch to GraphQL, rinse and repeat as needed

Phase 1 isn't strictly necessary, but I think it will help clean the code up a bit, aid in reuse, and also give us the flexibility to mix-and-match REST & GraphQL as necessary, among other benefits.

aorinevo:
I am definitely interested in optimizing calls and data transfer.  Before starting work on this, can you add a plan for delivering the update across all calls?  For example:

Batch | APIs | Target Date
---- | ---- | ----
1 | getPullRequestStatus, getCandidateRepos | 2101 B
2 | xxx, yy | 2102 A
... | ... | ...

Where 2101 B stands for the second half of January 2021, 2102 A stands for the first half of February 2021, and so on.

jugaltheshah:
> Strategy makes sense and I am definitely interested in optimizing calls and data transfer. Before starting work on this, can you add a plan for delivering the update across all calls? For example:
> Batch 	APIs 	Target Date
> 1 	getPullRequestStatus, getCandidateRepos 	2101 B
> 2 	xxx, yy 	2102 A
> ... 	... 	...
> 
> Where 2101 B stands for the second half of January 2021, 2102 A stands for the first half of February 2021, and so on.

This is a tough question to answer as it depends on a few major unknowns:
[...]

PR #377, opened 2021-01-21
feat: add github service

jugaltheshah:
### Issue 361 batch 1:
This PR starts work towards https://github.com/NerdWalletOSS/shepherd/issues/361 by adding a Github service. Idea here is to abstract the calls to Github behind something so that the underlying requests then can be changed out without doing too much work on the adapter. It will also allow us the flexibility to cleanly mix-and-match REST & GraphQL calls as needed.

#### Note
While all Github (more specifically, octokit) calls were moved to the new service, the abstraction is a bit leaky in that some adapter calls are still tied directly to Github responses. While not ideal, I think this is adequate for now  as this will be changing shortly with phase 2 and beyond. Overall, I believe this PR meets its objective of making the next phase(s) easier.


src/services/github.test.ts
```diff
@@ -0,0 +1,329 @@
+import type { Octokit } from '@octokit/rest';
+// import { RestEndpointMethodTypes } from '@octokit/plugin-rest-endpoint-methods'
```
jugaltheshah, line 2:
```suggestion
```

src/services/github.test.ts
```diff
@@ -315,15 +315,24 @@ describe('GithubService', () => {
...
         const service = new GithubService(mocktokit);
-        const criteria = {
+        const criteria1 = {
             search_type: 'code',
             search_query: 'org:testOrg path:/ filename:package.json in:path'
         };
 
-        const result = await service.getActiveReposForSearchTypeAndQuery(criteria);
+        const criteria2 = {
+            search_query: 'org:testOrg path:/ filename:package.json in:path'
```
aorinevo, line 324: Should one of these include a property `search_type` set to `""` or some falsey value?

jugaltheshah: We could, but it shouldn't be necessary. It's an optional parameter and when undefined, it hits the default case. In fact, if `search_type` is set to "" or a falsy value, the code will error due to https://github.com/NerdWalletOSS/shepherd/pull/377/files/5ad0285e1d7b7e12f1dc57cdf63ba23b82772902..982c25c27099f41d87a347f7a2345c81bf825aa6#diff-a9d5c72f5483cdb150974d030822fe95aeeba82f63bf72ae4b2c929fa7e50fb5R111. There's also a test that covers a bad value being passed: https://github.com/NerdWalletOSS/shepherd/pull/377/files/5ad0285e1d7b7e12f1dc57cdf63ba23b82772902..982c25c27099f41d87a347f7a2345c81bf825aa6#diff-8cde3a22acd42e1d42cdd48cb278b9902dce34a3f629dcb790841f2f8154f9b4R247.

aorinevo: Makes sense.

aorinevo:
PR looks good to me.  Agree that this PR meets the objective of phase 1.

aorinevo:
@jugaltheshah, just need conflicts resolved before merging.

aorinevo (approved):

aorinevo:
Did one last round of local testing.  Looks good.

PR #465, opened 2021-03-22
feat(commit): add "--no-verify" flag to "commit" command

davidburhans-domuso:
Then passes that option to the `git commit` command if present.


jderochervlk:
This is great!

jderochervlk (approved):

Issue #494, opened 2021-09-08
Travis CI workflow step pending for days


aorinevo:
## Summary

There are a number of PRs that are waiting to run on Travis; some on the order of days.  There are a couple of recommendations in [1] to resolve/troubleshoot the issue.  One recommendation is to parse and validate the Travis yaml file via [2].  The other recommendation is to check the requests in the Travis CI dashboard (see [3]).

I did not find any actionable feedback when inspecting the requests on the Travis CI dashboard but validating the yaml did result in some warn and info level messaging (see Appendix A).

## Next Steps
1. ~~[Interim] Open PR to update the Travis yaml to resolve the warn and info level messages.~~
2. [Long term] Move to GitHub Actions.

## Additional Items
1. ✅ Refresh GitHub token as the format has changed.  
2. ✅ Add NPM access token to GitHub Secrets.  This token is used to publish new tags to NPM via GitHub Actions.

### References
1. https://github.com/travis-ci/travis-ci/issues/7974
2. https://config.travis-ci.com/explore
3. https://travis-ci.org/github/NerdWalletOSS/shepherd/requests

### Appendix A
```
Parsed and validated config
[warn] on jobs.include.deploy: deprecated key: "skip_cleanup" (not supported in dpl v2, use cleanup)

[info] on jobs.include.deploy: the key api_key is an alias for api_token, using api_token

[info] on root: missing os, using the default "linux"

[info] on root: missing dist, using the default "xenial"
```

nwalters512:
👍 thanks for looking into this! It may be time to just move to GitHub Actions too.

aorinevo:
Happy to take a look at that as well as part of this.

aorinevo:
## Status Update

Tests are running via GitHub actions against Node version 12, 14, and 16.  Travis CI config runs test against 10, 12, and 14.  While I don't think we should add 10 to the test suite, I am open to adding it to maintain some parity with Travis CI configs in the short term (e.g., 1-2 weeks).  

Last step is to publish new tags to NPM.

aorinevo:
Refreshed GH token and NPM token added to GitHub Secrets.

aorinevo:
This issue is closed by https://github.com/NerdWalletOSS/shepherd/pull/499

PR #495, opened 2021-09-09
chore: add GitHub Actions.

aorinevo:
## Changes
- Adding GitHub action to handle testing against node 12, 14, and 16.

## References
1. see https://github.com/NerdWalletOSS/shepherd/issues/494


aorinevo:
This change is ok to merge to master because it only adds a GitHub action.  

Note Travis CI doesn't currently run.

PR #503, opened 2021-10-01
docs: Change the pr_message examples to use single-quotes

dbirks:
So the shell doesn't try to run the backtick-wrapped parts as commands.

To fix this essentially:
```
> echo "Hey! This PR renames `.eslintrc` to `.eslintrc.json`"
.eslintrc: command not found
.eslintrc.json: command not found
Hey! This PR renames  to
```

---

Also this is my first time using conventional-commits. Let me know if I messed anything up, or if I could do things better.


nwalters512 (approved):
Great catch! Thanks for contributing.

PR #504, opened 2021-10-01
fix: Make semantic-release commit the updated package.json files

dbirks:
I was wondering why the version was stuck at 1.14.2 in the package.json,
so I did some looking around. It looks like [^1] the order of plugins
matters, like the npm plugin was modifying the package.json and
package-lock.json after the git commit was already made.

[^1]: https://github.com/semantic-release/semantic-release/issues/1593#issuecomment-656866839


dbirks:
Oh and my employer gives me time to work on open-source for Hacktoberfest (part of the reason I can finally look for ways to contribute. I've been wanting to for a while because Shepherd is such a handy tool.) 

Would it be ok to label my PRs with `hacktoberfest-accepted`? (Or the whole repo could be labeled `hacktoberfest`, but then the repo would show up in searches for hacktoberfest -- pros and cons there.)

.releaserc.js
```diff
@@ -44,6 +44,7 @@ const releaseConfig = {
                 parserOpts
             }
         ],
+        '@semantic-release/npm',
```
aorinevo, line 47: The default setting for `@semantic-release/npm` is to also publish to our npm registry.  We should set the `npmPublish` configuration option to false.

The actual publishing to npm happens as part of the release github workflow.

dbirks: Oh ok, good catch. I'll try it out on my fork and update later this afternoon.

aorinevo:
Good catch!  Added a comment regarding the reordering.  Once that's resolved, I'd be happy to approve.

aorinevo:
> Oh and my employer gives me time to work on open-source for Hacktoberfest (part of the reason I can finally look for ways to contribute. I've been wanting to for a while because Shepherd is such a handy tool.)
> 
> Would it be ok to label my PRs with `hacktoberfest-accepted`? (Or the whole repo could be labeled `hacktoberfest`, but then the repo would show up in searches for hacktoberfest -- pros and cons there.)

I would be hesitant to add this to our github labels.  How about adding that to the PR title or description?

dbirks:
Oh that's right, the labels stick around and show in the label dropdown. 

Well it looks like the only other option is a repo topic. But it's also ok if these don't count as Hacktoberfest PRs. I still got to learn about Renovate and semantic-release. :slightly_smiling_face: 

Screenshot from 2021-10-01 10-04-50

Trying the tweaks in my fork now...

.releaserc.js
dbirks: So I tried setting `npmPublish` to false on my fork (in https://github.com/dbirks/shepherd/commit/bc604397e741e8f1f13cc353357c824b1b725f7f), but it didn't publish a package (no 1.15.6 version [here](https://github.com/dbirks/shepherd/packages/1019243)). 

Looking at the Actions in the release.yml, I don't see anything else that would publish the package, so I think it's been semantic-release that's been doing it.

Here's some examples of it publishing in my fork: the [Release Action](https://github.com/dbirks/shepherd/runs/3769772989?check_suite_focus=true), and the commit it makes with the changelog, package.json, and package-lock bumped (https://github.com/dbirks/shepherd/commit/32ba69651fdd125a8141236ff66216f3095ead5c).

Let me know if you're seeing something I'm not seeing. :mag:

aorinevo: Ah, right!

Then this should be fine.  I was mainly concerned that the version in Github would be behind the version in npm.

aorinevo (approved):

.releaserc.js
dbirks: Ok cool! :slightly_smiling_face:

PR #505, opened 2021-10-01
fix: Remove engines block from Renovate config

### LINKED ISSUE #480: Action Required: Fix Renovate Configuration
There is an error with this repository's Renovate configuration that needs to be fixed. As a precaution, Renovate will stop PRs until it is resolved.

Location: `renovate.json`
Error type: The renovate configuration file contains some invalid settings
Message: `The "node" object can only be configured at the top level of a config but was found inside "packageRules[2]", The "npm" object can only be configured at the top level of a config but was found inside "packageRules[2]"`

dbirks:
Looks like Renovate hit some sort of bug where it's moving the engines block down under packageRules for some reason. Here's what the resolved config looks like in the Renovate UI for my fork:

screenshot

I don't see engines configured in the package.json, and I'm not seeing a .nvmrc file or anything, so I _think_ you would be ok to just remove this engines block completely. It opened the Renovate flood gates when I was testing on my fork at least. 🌊

Fixes #480


aorinevo:
Nice!  Thank you for looking into this.

aorinevo (approved):

PR #506, opened 2021-10-01
fix: Remove space from .eslintignore filename

dbirks:
Just something small I noticed. Doesn't seem to make a difference when linting though, for some reason. Maybe one of the other eslint configs we're extending is excluding node_modules already. Not sure.


aorinevo (approved):

aorinevo:
Thanks for the contribution @dbirks!

PR #549, opened 2021-11-17
chore: Switch to Yarn

nwalters512:
This PR moves us to using Yarn for dependency management - it's what we use internally at NerdWallet and has generally proven to be a better experience than `npm`. The existence of tools like [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate) is especially helpful for solving obscure dependency issues.

"Obscure dependency issues" are actually what prompted me to make this change: I was trying to install https://www.npmjs.com/package/@octokit/plugin-throttling, but TS builds failed due to mismatched versions of Octokit types being installed. A `yarn import && yarn-deduplicate` fixed things right up!

The only part I'm not 100% sure about here is the semantic-release config - it's hard to test that until we actually merge to master. If it doesn't work as intended, we can fix forward!


.github/workflows/release.yml
```diff
@@ -9,20 +9,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
       - name: Setup Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
```
nwalters512, line 14: Figured I may as well bump these while I was here for consistency with the `ci.yml` flow.

.releaserc.js
```diff
@@ -51,7 +51,6 @@ const releaseConfig = {
             {
                 assets: [
                     'package.json',
-                    'package-lock.json',
```
FoodProduct, line 54: This is a bit of a shot in the dark since I'm not familiar with this config, but should this be replaced with `yarn.lock`?

.gitignore
```diff
@@ -4,68 +4,18 @@ lib/
 # Logs
 logs
 *.log
-npm-debug.log*
```
nwalters512, line 7: Started by removing this and ended up just cleaning up the entire file!

.gitignore
```diff
@@ -4,68 +4,18 @@ lib/
...
-jspm_packages/
-
-# TypeScript v1 declaration files
-typings/
-
-# Optional npm cache directory
-.npm
-
-# Optional eslint cache
-.eslintcache
```
FoodProduct, line 49: Is this not being generated, or are we purposely wanting to check this in, or other?

.gitignore
```diff
@@ -4,68 +4,18 @@ lib/
...
-
-# dotenv environment variables file
-.env
-
-# next.js build output
-.next
-
-# IDEs
-.idea
-*.iml
```
FoodProduct, line 68: Should we keep this in just in case anyone uses a JetBrains IDE?

FoodProduct (commented):
This generally makes sense to me -- I just have a few small q's!

.releaserc.js
nwalters512: I think the only reason this was here before was that `package-lock.json` actually contained the version for some reason. `yarn.lock` does not, so I don't believe we need it here.

.gitignore
nwalters512: We haven't opted in to caching in this repo, so it's unnecessary.

.gitignore
nwalters512: Probably, good call.

FoodProduct (approved):

aorinevo:
> A `yarn import && yarn-deduplicate` fixed things right up!

I'm not advocating for `npm` or `yarn` but I curious if running `npm dedupe`, or  `rm package-lock.json && npm install`, would have also resolved the issue.

> The only part I'm not 100% sure about here is the semantic-release config

I don't see a reason why `semantic-release` would not work with yarn.  My initial concern was in the publishing step but as yarn publishes to npm registry by default, I expect credentials and such to work as before.  Agree with the level-of-effort in testing `semantic-release` and understand that the risk can be mitigated with a roll forward strategy.

Changes look good to me.

aorinevo (approved):

nwalters512:
> I'm not advocating for npm or yarn but I curious if running npm dedupe, or rm package-lock.json && npm install, would have also resolved the issue.

`npm dedupe` is colossally bad in my experience - admittedly, I didn't try it in this scenario, but we found time and time again that it failed to dedupe trivial cases at NerdWallet, which is one of the things that ultimately led to our wholesale move to Yarn.

The latter would have likely worked, but I very much do not like blowing away lockfiles as a solution to our problems 😅

PR #551, opened 2021-11-18
feat: GitHub request throttling

nwalters512:
This PR adds [`@octokit/plugin-throttling`](https://github.com/octokit/plugin-throttling.js), which will complement [`@octokit/plugin-retry`](https://github.com/octokit/plugin-retry.js) that we already use.


src/services/github.ts
```diff
@@ -29,12 +30,16 @@ export default class GithubService {
         set a token on the 'password' field in ~/.netrc for api.github.com`);
       }
 
-      this.octokit = new RetryableOctokit({
+      this.octokit = new RetryableThrottledOctokit({
         auth: token,
-        retry: {
-          // By default the doNotRetry setting includes 403, which means we face secondary rate limits
-          doNotRetry: [400, 401, 404, 422],
-        }
```
nwalters512, line 37: I dropped this since the new plugin should now handle `403` errors for us.

aorinevo: Nice!

aorinevo (approved):

benjroy (approved):
Looks good to me

PR #323, opened 2020-12-02
feat: exclude archived repos when fetching repos by org

luislew:

PR #564, opened 2021-12-02
feat: Exclude archived repos when using code search

nwalters512:
Implements the same feature as https://github.com/NerdWalletOSS/shepherd/pull/323 but for code search!

### DIFF
--- a/src/services/github.ts
+++ b/src/services/github.ts
@@ -125,8 +125,7 @@ export default class GithubService {
   public async getActiveReposForSearchTypeAndQuery({ search_type, search_query }: SearchTypeAndQueryParams): 
   Promise<string[]> {
     if (search_type && !VALID_SEARCH_TYPES.includes(search_type)) {
-      throw new Error(`"search_type" must be one of the following:
-        ${VALID_SEARCH_TYPES.map(e => `'${e}'`).join(' | ')}`);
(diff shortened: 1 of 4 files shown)


src/services/github.ts
```diff
@@ -61,14 +61,12 @@ export default class GithubService {
...
-    await this.paginateRest(this.octokit.search.code, criteria);
-
-    return searchResults.map((r) => _.get(r, 'repository.full_name')).sort();
+  private async findReposByCode(criteria: RestEndpointMethodTypes['search']['code']['parameters']):
+  Promise<RestEndpointMethodTypes['search']['code']['response']['data']['items']> {
+    return this.paginateRest(this.octokit.search.code, criteria);
   }
 
-  private getRepos(criteria: RestEndpointMethodTypes['repos']['get']['parameters']):
+  private getRepo(criteria: RestEndpointMethodTypes['repos']['get']['parameters']):
```
nwalters512, line 69: AFAICT this only gets a single repo, so this name seemed more fitting.

aorinevo: Checked the Octokit docs and it seems that's the case (see [1]).

References
1. https://octokit.github.io/rest.js/v18#repos-get

aorinevo (approved):

aorinevo (approved):

PR #565, opened 2021-12-02
fix: Use custom GitHub token to bypass branch protection

nwalters512:
Our few most recent commits to master have failed to publish because `semantic-release` couldn't push new versions to `master`. Per https://github.com/semantic-release/semantic-release/blob/8fda7fd423d24e7b425fbee83790f49dd0478e2d/docs/recipes/ci-configurations/github-actions.md#pushing-packagejson-changes-to-a-master-branch, we should be setting `persist-credentials: false` during checkout so that our custom token (that can bypass branch protection restrictions) is used by `semantic-release`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


aorinevo (approved):

nwalters512:
Looks like this worked - v1.16.0 was just published!

PR #572, opened 2021-12-21
Adds prettier dependencies and configuration

scalvert:
This PR is split into two commits.

Commit 1:

Adds `prettier` dependencies and configuration:

- `.prettierrc.js`
- `.prettierignore`
- `prettier` and `eslint-config-prettier` dependencies
- `eslint` configuration

Commit 2:

Runs `npx prettier --write .` to reformat all the files.


package.json
```diff
@@ -76,9 +76,11 @@
     "@typescript-eslint/eslint-plugin": "^4.15.0",
     "@typescript-eslint/parser": "^4.15.0",
     "eslint": "^7.19.0",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-import": "^2.22.0",
     "eslint-plugin-prefer-arrow": "^1.2.3",
     "jest": "^27.0.0",
+    "prettier": "^2.5.1",
```
nwalters512, line 83: Let's lock Prettier to an exact version like they recommend in the [installation docs](https://prettier.io/docs/en/install.html).

nwalters512 (changes requested):
Let's make sure that that `lint` and `fix-lint` commands verify and fix formatting, respectively - that'll allow us to verify formatting in CI. AFAICT, the current CI workflow won't actually fail if things aren't formatted correctly.

scalvert:
Thanks for the comments. I think `yarn lint` will in fact fail CI if there are prettier issues, via the eslint configuration. We probably shouldn't auto-fix using `--fix` in CI, if that's what you're suggesting. What do you think?

package.json
scalvert: Makes sense. I'll update it.

nwalters512:
`extends: ['prettier']` in the ESLint config only disables rules that overlap with what Prettier does, it doesn't actually enforce formatting. I tested this by manually mis-formatting a file and then running `yarn lint` - it succeeded. You'd have to use something like https://github.com/prettier/eslint-plugin-prettier to actually enforce when running Prettier, though they explicitly recommend against that [in the docs](https://prettier.io/docs/en/integrating-with-linters.html). You'll need to add an invocation of Prettier to both the `lint` and `fix-lint` commands.

package.json
scalvert: Done.

scalvert:
Hmmm ya I suppose I usually run with `eslint-plugin-prettier`, though they suggest using `prettier --check .` now. I'll do that.

scalvert:
OK, fixed it up. I changed the `fix-lint` command to not call `yarn lint` in order to avoid running prettier twice.

package.json
```diff
@@ -23,8 +23,8 @@
   "scripts": {
     "build": "tsc",
     "build:watch": "yarn build --watch",
-    "fix-lint": "yarn lint --fix",
-    "lint": "eslint src/**/*.ts",
+    "fix-lint": "prettier --write . && eslint src/**/*.ts --fix",
+    "lint": "prettier --check . && eslint src/**/*.ts",
```
nwalters512, line 27: I generally prefer to run Prettier after ESLint so that it'll reformat any files that `eslint --fix` changed. Otherwise, you may have to run `fix-lint` twice.

scalvert: Makes sense. I'll fix it.

scalvert: I reversed both, for consistency.

scalvert:
Hey @nwalters512, when you get back from holidays let me know if there's anything else you'd like me to add to this.

nwalters512 (approved):
This looks great!

PR #573, opened 2021-12-21
chore: Adds cli test infra to enabling testing bin invocation

scalvert:
This PR adds basic bootstrapping of infra that will allow for CLI tests that directly invoke the bin. This can provide a path to end to end tests that will perform a complete migration.

Added:
- `execa` devDependency for invoking the CLI (I know this project uses child-process, but I've found execa to be really nice for this use case. Happy to discuss whether or not to add this)
- `fixturify-project` devDependency for building out ad-hoc migration project directories
- A simple example of a cli test using a `shepherd` util that maps to the bin script, and tests the usage information is correctly printed

The tests are in the standard `__tests__` jest directory, mainly because I wasn't sure where to add tests that weren't specifically unit tests (I suppose they could live next to the file itself, as the other tests do).

TODO
- [ ] Add a fixturify project subclass to provide utilities for adding a `shepherd.yml` to the migration directory
- [ ] Add the ability to mimic repos to migrate (separate fixturify project directories?)
- [ ] Add end to end example of testing a migration, from repo checkout to pr creation
- [ ] Exploring using something like https://mswjs.io/ to provide adapter mocking.


__tests__/cli.test.ts
```diff
@@ -0,0 +1,84 @@
+import execa from 'execa';
+import { Project } from 'fixturify-project';
```
nwalters512, line 2: 🤯 I didn't know something like this existed! This is awesome. I've done this by hand more times than I can count...

package.json
```diff
@@ -78,11 +78,16 @@
...
     "eslint-plugin-prefer-arrow": "^1.2.3",
+    "execa": "^5.0.0",
+    "fixturify-project": "^4.0.2",
     "jest": "^27.0.0",
     "semantic-release": "^17.3.8",
     "ts-jest": "^27.0.0",
     "typescript": "^3.9.9"
   },
+  "resolutions": {
+    "resolve-package-path": "^4.0.3"
```
nwalters512, line 89: Just out of curiosity, why is this necessary?

nwalters512 (approved):

__tests__/cli.test.ts
scalvert: We use it extensively, in countless projects. It's very powerful and allows you to do ad-hoc testing really easily.

package.json
scalvert: There was a bug in version 3 of `resolve-package-path` that I [fixed](https://github.com/stefanpenner/resolve-package-path/pull/38). That fix has not made it into `fixturify-project` yet, but is largely safe since it's just a change to the files that were published. I'm opening a PR right now to roll that into `fixturify-project`.

scalvert:
I've been letting this percolate over the holidays. I think I have a nice setup that will allow us to be able to create ad hoc migrations and target repos, and will also allow us to run full migration tests across those. I also think it'd be nice to provide test infra for migration authors to validate their migrations will run correctly. In that sense, decoupling the utils that allow you to generate fake target repos to run the migrations against would be useful to provide publicly to consumers.

@nwalters512 - what do you think?

nwalters512:
> I also think it'd be nice to provide test infra for migration authors to validate their migrations will run correctly. In that sense, decoupling the utils that allow you to generate fake target repos to run the migrations against would be useful to provide publicly to consumers.

Can you share a little more of what you have in mind? This *could* be useful, but I'd want to hear more details first.

scalvert:
@nwalters512 let me work on getting test infra setup that allows us to generate target repos in our CLI tests. Once I've taken a stab at that, we can see if we like it and if we want to extend it past this PR (in a separate PR, of course). Sound good?

scalvert:
So I'm going to continue the work on this PR (sorry for the delay!). I'm going to narrow the focus of adding some functionality to simply test dynamically creating a migration and a target repo, and running through those tests.

Subsequent PRs can add public test helper support to exercise testing a migration for migration authors.

PR #586, opened 2022-04-07
feature, add ability to skip pre-commit/pre-push with --no-verify

Mechanicalkeyboard:


aorinevo:
IMO, these parameters should not be passed through Shephard as they are typically, amongst other things, used to ensure certain standards are being met before committing or pushing changes which can help save on, amongst other things, resource consumption downstream (e.g., build pipelines, integ tests, etc...).

One workaround to consider is to modify your local git config via
```
git config --global core.hooksPath /dev/null
```
You can revert back using
```
git config --global --unset core.hooksPath
```

Thoughts?

Mechanicalkeyboard:
@aorinevo I do like your suggestion of modifying the global hooks locally for certain cases. 

I would like the option in this repo, because when you are working against many flavors of repositories, having the option of skipping pre-commit/pre-push hooks can save time in getting small isolated changes out to many repos where they may have hooks that trigger tests that require local setup. 
My use case for this I end up running shepherd for hours, against many repos, modifying my global git config would interrupt my regular development life cycle. I'm basically pushing the heavy ci tests/ config require tests to our ci platform.

AndrewCsontos:
> 

Being able to add --noverify would be a really useful addition for my teams as well.

PR #605, opened 2023-09-11
BREAKING CHANGE: dropping support for eof node versions 12, 14 and 16; upgrade dependencies

aorinevo:
# Changes
* Update CI dropping support for EOL Node versions 12, 14, and 16. 
* Bump Release Node version to 18, latest supported Node version.


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   },
   rules: {
     '@typescript-eslint/no-unsafe-assignment': 'off',
+    '@typescript-eslint/no-unsafe-argument': 'off',
```
aorinevo, line 20: To handle the pervasive use of implicit any everywhere.  We should follow-up with proper typing but out-of-scope of this PR as it is an existing issue and scope is narrowed down to dependency management at this time.

.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 12
+          node-version: 18
```
aorinevo, line 22: Bumping to minimum supported version.

.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['12', '14', '16']
+        node: ['18', '20']
```
aorinevo, line 8: As mentioned in the description, versions < 18 are EOL as of 9/11/2023.

src/util/migration-spec.ts
```diff
@@ -26,7 +26,7 @@ export interface IMigrationSpec {
 
 export function loadSpec(directory: string): IMigrationSpec {
   const docPath = path.join(directory, 'shepherd.yml');
-  const spec = yaml.safeLoad(fs.readFileSync(docPath, 'utf8'));
+  const spec = yaml.load(fs.readFileSync(docPath, 'utf8'));
```
aorinevo, line 29: Yaml package renamed safeLoad to load.  In this PR, we update to the new name.

src/util/persisted-data.ts
```diff
@@ -16,7 +16,7 @@ const jsonStringify = (data: any) => JSON.stringify(data, undefined, 2);
 const migrateToJsonIfNeeded = async (migrationContext: IMigrationContext) => {
   const legacyFile = getLegacyRepoListFile(migrationContext);
   if (await fs.pathExists(legacyFile)) {
-    const data = yaml.safeLoad(await fs.readFile(legacyFile, 'utf8'));
+    const data = yaml.load(await fs.readFile(legacyFile, 'utf8'));
```
aorinevo, line 19: Same as previous comment.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nerdwallet/shepherd",
-  "version": "1.16.0",
+  "version": "2.0.0",
```
aorinevo, line 3: This manual bump is unnecessary as semantic-release will bump the major version.

src/logger/index.ts
```diff
@@ -58,28 +56,34 @@ export default class ConsoleLogger implements ILogger {
...
 
   public spinner = (message: string): ISpinner => {
+    const that = this;
     if (this.oraInstance) {
       this.oraInstance.stop();
     }
-    this.oraInstance = ora(message).start();
-    this.spinnerActive = true;
-    this.oraInstance.start();
+    import('ora')
```
aorinevo, line 79: This dynamic import is a workaround and not a good one.  It causes race conditions that are more challenging to solve than to convert to using a different spinner.  Need to evaluate alternatives.

src/logger/index.ts
```diff
@@ -58,28 +56,34 @@ export default class ConsoleLogger implements ILogger {
   };
 
   public succeedIcon = (message: string): void => {
-    this.info(`${logSymbols.success} ${message}`);
+    this.info(`✅ ${message}`);
```
aorinevo, line 59: One option to removing dependency on log-symbols which, as in the case of ora, requires dynamic imports.  I'm fine with using emoji, unless there's a preference or reason to do otherwise.

aorinevo:
Almost there.  Local testing looks good but there are a few issues left outstanding that need to be resolved:
1. Deprecate use of log-symbols in favor of similar TS module or just use emoji
2. Deprecate ora spinner in favor of similar TS module

I'm leaning towards deprecation of both log-symbols and ora.  Open to feedback :)

.eslintrc.js
```diff
@@ -16,17 +16,19 @@ module.exports = {
     sourceType: 'module',
   },
   rules: {
-    '@typescript-eslint/no-unsafe-assignment': 'off',
```
aorinevo, line 19: Prefer integer value.  0 = off, 1 = warn, and 2 = error.

PR #629, opened 2023-11-18
feat: added options to create pr from fork to upstream

kavitha186:
## Background

This PR adds a feature to enable use of Shepherd to raise a PR from a fork to upstream.  The feature is exposed via the `--upstreamOwner` CLI option.  When present with non-null value in the `pr` command, the PRs will be opened towards `${upstreamOwner}:{branchName}` for all in-scope repos.

It's not immediately obvious and so worth highlighting that the Shepherd `push` command does not require any changes to work as expected.  Just note that pushing changes is done relative to the remote and branch configuration as set during Shepherd checkout.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -45,6 +45,7 @@ type CommandHandler = (context: IMigrationContext, options: any) => Promise<void
 
 interface ICliOptions {
   repos?: string[];
+  upstreamOwner: string;
 }
 
 const handleCommand =
@@ -76,6 +77,7 @@ const handleCommand =
 
       // The list of repos will be null if migration hasn't started yet
       migrationContext.migration.repos = await loadRepoList(migrationContext);
+      migrationContext.migration.upstreamOwner = options.upstreamOwner;
 
       await handler(migrationContext, options);
     } catch (e: any) {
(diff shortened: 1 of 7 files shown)


src/cli.ts
```diff
@@ -94,10 +96,15 @@ const addReposOption = (command: program.Command) => {
   );
 };
 
+const addUpstreamOwnerOption = (command: program.Command) => {
+  return command.option('--upstreamOwner <upstreamOwner>', 'Upstream Owner can be passed incase of trying to raise PR from fork to upstream');
```
aorinevo, line 100: Perhaps this option should be called forkOwner instead of upstreamOwner to provide more clarity on its use.  When forkOwner is defined, PR will be generated off of fork to upstream instead of creating a branch on upstream.

aorinevo:
Updated the PR description.  One thing that I'm not entirely sure of is whether or not we should, in addition to specifying the upstreamOwner, also specify the upstream branch.  In lieu of that, fork branch name and upstream branch name must be the same, and that's not always guaranteed.

aorinevo (commented):
See prior comments.

kavitha186:
> Updated the PR description. One thing that I'm not entirely sure of is whether or not we should, in addition to specifying the upstreamOwner, also specify the upstream branch. In lieu of that, fork branch name and upstream branch name must be the same, and that's not always guaranteed.
> 
> Thoughts?

The PR command creates PR from pushed branch on fork to upstream repo master/default branch. Do you think is it necessary to specify the upstream branch?

aorinevo:
Ah, if Shepherd pr command will use the default branch, then I don’t have a concern, however, there might be value in allowing users to specify the upstream branch that’s not default - not a blocker for this PR.

aorinevo:
FYI, I'll cherry-pick this commit into a feature branch on upstream and open a PR.  Once that is reviewed, approved, and merged, this PR should close automatically.

aorinevo:
@kavitha186, in the process of cherry-picking this over to an upstream branch, CI fails on linting.  Can you fix the linting issues and push updates to this branch?   If you have multiple commits, please squash them into 1 before updating your branch.

aorinevo:
Closed in favor of https://github.com/NerdWalletOSS/shepherd/pull/633

Issue #609, opened 2023-09-16
Add support for GitHub Enterprise


aorinevo:
# Changes
1. Add support for GitHub Enterprise by making the GitHub API configurable.
2. Url should default to api.github.com.
3. Simple Git needs to be configurable to use enterprise url as well.

aorinevo:
Closed by #631

aorinevo:
Closed by #631

PR #631, opened 2024-01-07
feat(gh-enterprise): Add support for github enterprise servers

aorinevo:
## Background

This PR adds support for GitHub Enterprise Servers. 

Currently, Shepherd assumes that repos can be pulled, pushed, etc... via `api.github.com`.  This assumption fails to be true when working with GitHub Enterprise Servers which have different domain names.  Support for GitHub Enterprise Servers is established by leveraging the `SHEPHERD_GITHUB_ENTERPRISE_URL` environment variable to determine the correct domain name.

This PR also takes care of some routine dependency management.


aorinevo:
Closes https://github.com/NerdWalletOSS/shepherd/issues/609

kavitha186 (approved):

Issue #87, opened 2019-12-07
Using shepherd to file issues?


TheSavior:
Hi! I'm on the React Native team at Facebook. We are working on a big reimplementation of the core framework which will result in a few different breaking changes in the API. Unfortunately, not all of these changes can be migrated with codemods. If they could, then shepherd would be an absolutely perfect tool for us to use to help migrate the community.

Instead, for some of these breaking changes, we are planning on writing guides and instructions for how to make these changes. Even those this work can't be automated, we'd like to help the community migrate as easily as possible. 

I think Shepherd has a lot of the pieces to help us with this!

What I would love to do is file issues on the repos that match the search criteria, linking to the lines of code that needs to change.

I can definitely imagine that this is out of scope of the purpose of shepherd, but a lot of the basics seem consistent.

I'm curious if you have any thoughts for us, or if you'd recommend just building something separate?

TheSavior:
From looking through the code, it seems like a reasonable approach would be to implement commands for issues, similar to the ones for PRs:
 https://github.com/NerdWalletOSS/shepherd/blob/538f65ff4764cdd290b296c350821374181c06d4/src/cli.ts#L19-L21

And then we'd run only these commands:
```
shepherd checkout ~/shepherd-migration
shepherd issue ~/shepherd-migration
```

Does that sound right to you?

parshap:
Hey @TheSavior. This is awesome! Using Shepherd to lessen the burden of api changes is definitely type of use case we want to support.

Your idea with `shepherd issue` makes sense. I think the `shepherd.yml` migration file should also indicate that an issue should be opened instead of a pr (perhaps by using `issue_message` instead of `pr_message`). I would happily accept a PR for this.

PS: I've run into a similar use case before where I've wanted to add line comments on the PR that Shepherd opens.

TheSavior:
Great! Having the issue command read something like issue_message sounds reasonable to me

aorinevo:
Hi @TheSavior, apologies on the long hiatus.  I'm just getting back into the swing of things and wanted to revisit this.  I'm on the fence with respect to this feature.  On the one hand, Shepherd is primarily focused on programmatic code changes and so this would be outside it's scope.  On the other hand, the proposed feature is targeting reducing developer toil at scale by way of programmatically creating issues.

Also, for full transparency, I've had a similar use case and opt'd to add the feature to a custom cli tooling my team uses.  

[...]

Issue #590, opened 2022-06-06
Add support for labels


ghost:
It would be handy to be able to add labels (where supported) to PRs as part of this process.

sdann:
This looks to be a duplicate of issue https://github.com/NerdWalletOSS/shepherd/issues/54

aorinevo:
> This looks to be a duplicate of issue https://github.com/NerdWalletOSS/shepherd/issues/54

Yep, these seem to be the same.

aorinevo:
Closing in favor of #54

PR #663, opened 2024-01-20
feat: create issues with message and tag labels

kavitha186:
This PR is for the feature to post issues on the repositories with shepherd. if you are looking to post issues across multiple repositories, you can use the command "issue" to create, update and close issue

issue with the title, issue message, labels , state and state_reason provided in the migration scripts. 

This enhancement feature is for the issue #87


README.md
```diff
@@ -94,8 +94,9 @@ The options under `hooks` specify the meat of a migration. They tell Shepherd ho
...
 - `pr_message` is a sequence of commands that will be used to generate a pull request message for a repository. In the simplest case, this can just be a static message, but you could also programmatically generate a message that calls out particular things that might need human attention. Anything written to `stdout` will be used for the message. If multiple commands are specified, the output from each one will be concatenated together.
-
-`should_migrate` and `post_checkout` are optional; `apply` and `pr_message` are required.
+- `issue_message` is a sequence of commands that will be used to post a issue message for a repository. In the simplest case, this can just be a static message.
+- `issue_labels` is a sequence of commands that will be used to add labels to the issue posted to the repository.
+  `should_migrate` and `post_checkout` are optional; `apply` and `pr_message` are required. `issue_message` and `issue_labels` are required for `create-issue` command.
```
aorinevo, line 99: I'm thinking we can move this into a Schema section of the readme where it can be quickly referenced for required vs optional keys

aorinevo:
This PR also takes care of Issue #590

src/adapters/github.test.ts
```diff
@@ -11,6 +11,10 @@ const mockMigrationContext = () =>
       spec: {
         id: 'test-migration',
         title: 'Test migration',
```
aorinevo, line 13: nit: The value is used for the commit message and so we should use semantic commit convention (e.g., feat, docs, etc...)

src/commands/issue.ts
```diff
@@ -0,0 +1,25 @@
+import { IMigrationContext } from '../migration-context';
+import forEachRepo from '../util/for-each-repo';
+
+export default async (context: IMigrationContext) => {
+  const {
+    migration: { spec },
+    logger,
+  } = context;
+
+  if (!spec.hooks.issue_message || !spec.hooks.issue_labels) {
```
aorinevo, line 10: What's the rational behind making labels required?

src/commands/issue.ts
```diff
@@ -0,0 +1,25 @@
+import { IMigrationContext } from '../migration-context';
+import forEachRepo from '../util/for-each-repo';
+
+export default async (context: IMigrationContext) => {
```
aorinevo, line 4: If we open an issue and want to update it later, how would we go about doing that?  A basic use case would be a typo but can be more substantial, for example additional references or detail that was not initially captured when the issue was first published.

Similarly, I'm thinking labels should be updateable.

aorinevo:
cc @TheSavior for visibility.

src/commands/issue.ts
kavitha186: Yes, I will add another feature on top of this to get the issue with issue # and update it.

src/adapters/github.test.ts
kavitha186: I think the issue message doesn't need to have semantic convention.

src/commands/issue.ts
kavitha186: Should I make both issue message and labels as optional

aorinevo: It depends on the rational for making them required.  Sometimes Issues are opened with just a title; I can see title being required.

src/adapters/github.test.ts
aorinevo: Ah, I got my wires crossed 🥹

src/commands/issue.ts
kavitha186: yes i made both issue message and labels as optional

src/commands/issue.ts
aorinevo: I’m curious to learn more about the implementation details for this.  How are we thinking of persisting state so that we can update issues and labels?

README.md
kavitha186: yes I will refine the documentation in the next PR. Would you want to open an issue?

src/commands/issue.ts
kavitha186: the issue can be updated with issue # and it is challenging to track the issue number for multiple repositories. I will think of different way to persist the state

aorinevo: Take a look at Shepherd's existing persistence before rolling your own (unless, of course, there's a need).

README.md
aorinevo: The documentation should accompany the changes.  Happy to contribute towards that end.

aorinevo:
Created feature branch feat/create-issues-labels on upstream and pointed this pr to it.  Once the feature branch is ready to ship, we'll merge it down to main.

aorinevo:
Here's an in-progress PR to update the README: https://github.com/NerdWalletOSS/shepherd/pull/666

src/commands/issue.ts
aorinevo: Persisted-data tracks checkout repos under repos.json.  We can use a similar strategy to track issues, perhaps under an issues.json.  When an issue is created, using the GitHub Create Issues API, on success, we get a response that contains the Issue number.  We can track this number in a map of repos.  For example:

```
[
repo1: {
   issueNumber: 1
},
repo2: {
   issueNumber: 3
},
repo3: {
   issueNumber: 1
}
]
```

Migrations may not necessarily be 1:1 with issue creation.  That is, a migration can create more than one issue.  So the above won't suffice.  To accomodate multiple issues, we can store the data as follows:

```
[
repo1: {
  issues: [
     {
        issueNumber: 1, 
        issueKey: identifier1 // from migration spec; each issue will have a unique identifier
      }
   ]
},
repo2: {
   issues: [
     {
         issueNumber: 3, 
         issueKey: identifier2
     }
  ]
},
repo3: {
   issues: [
      {
         issueNumber: 1, 
         issueKey: identifier1
      },
      {
         issueNumber: 3, 
         issueKey: identifier2
      }
    ]
}
]
```

Persisting the data in this way will enable us to update issues as needed.

aorinevo: This is just a suggestion and I haven't considered alternative data structures and so this might not be optimal.  Note, I'm also sure we can add to the repos.json file but I haven't assessed whether that's going to benefit us in the long term or not.  Please evaluate accordingly.

aorinevo:
Should we only handle Create and Update?  What about Delete and Read?

src/commands/issue.ts
kavitha186: yes i also had a same thought however, the create and update issue should be a seperate flow and we can match the title and fetch the issue number to update the same issue

aorinevo: Why wouldn’t we be able to update the title?

TheSavior:
> Should we only handle Create and Update? What about Delete and Read?

What would Read do? I assume you'd Read them via GitHub notifications.
I could imagine wanting to close issues. For example if you send out bulk issues, and then realize you shouldn't have opened half of them. But that might be beyond the scope of this tool. I'm not sure Shepherd gives you a way to do that with PRs either.

Fwiw, the way I was planning to use this, and is maybe a good recommendation for how everyone should use this, is having 1 primary issue on some repository (in our case React Native), and then filing issues against all the relevant repos with a backlink to the primary issue. That would enable tracking all the sub issues and enabling repo maintainers to have a place to communicate and solve their problems together.

Just sharing that workflow here in case that is important for how it is implemented?

src/commands/issue.ts
kavitha186: Actually,  if the issue identifier with issue number can be persisted while creating an issues across multiple repos, the issue identifier can be repository name along with issue #- then if i post another set of issues to same set of repos, the issues.json will be overwritten, with new issue numbers. In that case, i can update the title, message, and labels. 

Solution#1: Save the state of posted issue with title + repo + issue# 

In this case, the title can't be updated, and more than 1 issue can be posted/updated on a repo, and each issue can be persisted in an issues.json

Solution#2: 

Persist the state of posted issues with repo name + issue #

In this case, the title can be updated, but only 1 issue can be persisted for a repo.

TheSavior: Ah, a likely thing I could imagine wanting (that is relevant to persisting), is adding an additional comment to the issue that was created. 

For example, the initial issue might be about some API they need to change. I might want to follow that up with a "There is 30 more days until this API is removed" or something like that.

kavitha186: We can have separate workflows for adding comments on the open issue. The lifecycle for posting comments on issue workflow should continue the "create issue" workflow. 

Create issue -> save the state of posted issues -> Read/update/delete issues & post comment on issue.

README.md
```diff
@@ -165,6 +168,9 @@ There are a number of commands that must be run to execute a migration:
 - `pr-preview`: Prints the commit message that would be used for each repository without actually creating a PR; uses the `pr_message` hook.
 - `pr`: Creates a PR for each repo with the message generated from the `pr_message` hook.
 - `version`: Prints Shepherd version
+- `create-issue`: Creates issue on the repo with `title`, `issue_message` and `issues_labels`.
```
aorinevo, line 171: What's the rationale for splitting out `create-issue` and `update-issue` vs. just having a single issue api.  For the latter, if an issue doesn't currently exist, we create it.  Otherwise, we update it.

src/adapters/base.ts
```diff
@@ -40,6 +44,12 @@ interface IRepoAdapter {
   getBaseBranch(repo: IRepo): string;
 
   getEnvironmentVariables(repo: IRepo): Promise<IEnvironmentVariables>;
+
+  createIssue(repo: IRepo): Promise<string>;
```
aorinevo, line 48: Consider exposing issue interface with repo as required and issueNumber as optional.  If issueNumber exists, update.  Otherwise, create a new issue.

src/adapters/base.ts
```diff
@@ -40,6 +44,12 @@ interface IRepoAdapter {
...
+
+  updateIssue(repo: IRepo, issueNumber: number): Promise<void>;
+
+  closeIssue(repo: IRepo, issueNumber: number): Promise<void>;
```
aorinevo, line 52: Similarly, I'm thinking this can be passed as a parameter to an issue api.

kavitha186: I can merge create and update as an issue api. But for close, can i use the same issue api if the state is passed as closed in the migration script

aorinevo:
Added some comments but thought I thought I would also summarize my thinking here.  While I'm not opposed to exposing more top level apis in Shepherd, I would like to understand the rational behind them better.  For example, why expose  `create-issue`, `update-issue`, and `close-issue` apis vs exposing a single `issue` api. 

As a developer, I would prefer to run one command that does all the heavy lifting for me versus having to orchestrate/run 3 separate commands.  Shepherd is able to, by persisting state, take a decision on whether to update, create new, or close an issue based on the current configuration.

src/adapters/base.ts
aorinevo: I think so.  My guess is that there'll be some version of the following logic in the code:

```
if(issueExists){
   if(stateIsClosed){
      // call manage state function
   }  else {
      // call update issue function
   }
} else {
   // call create issue function
}
```

kavitha186: i merged update and close

src/commands/issue.ts
aorinevo: If we choose to support comments, we would certainly need a reference to the issues on which to comment.  We would likely need to maintain state and provide a view of open issues, their state, and repo.  

This use case highlights a more general feature gap wherein developers should have view access to Shepherd's state.  This would help them better understand what actions to take through Shepherd (e.g., list PRs associated with a migration and their state).

cc @TheSavior

README.md
```diff
@@ -81,6 +81,10 @@ hooks:
   post_checkout: npm install
   apply: mv .eslintrc .eslintrc.json
   pr_message: echo 'Hey! This PR renames `.eslintrc` to `.eslintrc.json`'
+  issue_message: 'This is the test message'
```
aorinevo, line 84: As issue_message, issue_labels, state_reason, and state are all associated with issues, I think it makes sense to store these properties in an object set on an issue key:

```yaml
issue:
  message: 'This is the test message'
  labels: ['bug', 'enchancements']
  state_reason: ['completed']
  state: 'closed'
```

Questions:
1. Should we have a title property?  This would allow for a title different from the spec title, which is used for the commit message.
2.  Does message map to issue description?  If so, thoughts on naming it description vs. message?
3.  What is state_reason?  Is this the comment used to close the issue?
4. Thoughts on making state a enum?

src/adapters/base.ts
```diff
@@ -40,6 +44,10 @@ interface IRepoAdapter {
   getBaseBranch(repo: IRepo): string;
 
   getEnvironmentVariables(repo: IRepo): Promise<IEnvironmentVariables>;
+
+  createIssue(repo: IRepo): Promise<string>;
```
aorinevo, line 48: What's the rational for exposing createIssue and updateIssue vs createOrUpdateIssue?

aorinevo:
Can we also add a command to list out the issues?  For example, `shepherd list-issues .` would return

```
issue number | issue title | issue state
---- | ---- | ----
1 | This is an issue associated with the migration | Open
2 | This is another issue associated with the migration | Closed
```

This will enable developers to sanity check the migration from the command line vs having to navigate to GitHub to verify.
Thoughts?

src/adapters/base.ts
kavitha186: The github api is different for create abd update so I feel adapter interface should separate the calls to different api.

kavitha186:
> Can we also add a command to list out the issues? For example, `shepherd list-issues .` would return
> 
> issue number	issue title	issue state
> 1	This is an issue associated with the migration	Open
> 2	This is another issue associated with the migration	Closed
> This will enable developers to sanity check the migration from the command line vs having to navigate to GitHub to verify. Thoughts?


@aorinevo 
Questions:
1. Should we list the issues on a repo/ list of issues opened/closed with migration? 
2. Is it viewable if the list is long?

src/migration-context.ts
```diff
@@ -21,3 +21,30 @@ export interface IMigrationContext {
   adapter: IRepoAdapter;
   logger: ILogger;
 }
+
+export enum labels {
```
aorinevo, line 25: How would we go about supporting custom labels if we use enum?

src/util/migration-spec.ts
```diff
@@ -3,15 +3,25 @@ import fs from 'fs';
 import * as yaml from 'js-yaml';
 import { cloneDeep, mapValues } from 'lodash';
 import path from 'path';
+import { state, state_reason } from '../migration-context';
 
 export interface IMigrationHooks {
   should_migrate?: string[];
   post_checkout?: string[];
-  apply?: string[];
+  ply?: string[];
```
aorinevo, line 11: Should this be apply?

Issue #670, opened 2024-01-23
GitHub Adapter/Service Updates


taylorperkins:
Hey hey! Thank you all for putting this together. Very cool project! I am not a JS/TS dev, per se, so please be patient with me ❤️ However, I am happy to help/contribute if possible. My team at [iheartradio](https://github.com/iheartradio/) is considering a tool like this, and so far sheperd does pretty much everything we would want! However, we noticed a couple of things that may need to be tweaked (at least for our use-case). Could definitely be user-error though!

I started playing around with shepherd yesterday (great tutorial btw!) and noticed a couple of things that were not working out. 

## Context
* iheartradio is a part of the [iheart-media-inc](https://github.com/enterprises/iheart-media-inc/) enterprise
* iheartradio would be using the GitHub service/adapter
* The contents of the `shepherd.yml` is: 
```yaml
id: 2024.01.22-sync-files
title: Sync all watched files
adapter:
  type: github
  search_type: repositories
  search_query: org:iheartradio topic:mle
hooks:
  should_migrate:
    - ls "$SHEPHERD_MIGRATION_DIR/watchers" | grep $(basename "$PWD") # Check that the repo has a watch configuration
  apply: "${SHEPHERD_MIGRATION_DIR}/sync.sh"
  pr_message: echo 'Hey! This PR syncs all files'
```

So - we are trying to apply the `sync.sh` script to all repositories in the `iheartradio` org that contains the `mle` topic. 

I was able to get this working for us using this yaml config, just not at first.

## Checkout Search Issue
When trying out the first step, `checkout`, I was not able to get it to work using the enterprise URL. I believe this errors out when trying to query for repositories based on the `shepherd.yml`. 

For example:
```
export SHEPHERD_GITHUB_ENTERPRISE_URL=api.github.com/enterprises/iheart-media-inc
```
Which produces:
```
RequestError [HttpError]: Not Found
...
{
    url: 'https://api.github.com/enterprises/iheart-media-inc/api/v3/search/repositories?q=org%3Aiheartradio%20topic%3Amle',
    status: 404,
    headers: {
      'access-control-allow-origin': '*',
[...]

PR #674, opened 2024-01-24
Issue 670/gh adapter service updates

taylorperkins:
## References [Issue 670](https://github.com/NerdWalletOSS/shepherd/issues/670)
See the issue for a full detail on why this change was made. 

This may not be all of the place that these changes need to be made, and they may not need to be made. These changes were made after going through the instructions in the README (specifically for enterprise), and not being able to go through the full shepherd lifecycle, specifically `checkout`. After making these changes, I can go from `checkout` -> `pr` for the iheartradio organization. 

If there is something else I can try using the originally implementation, I am happy to!

### DIFF
--- a/src/adapters/github.ts
+++ b/src/adapters/github.ts
@@ -309,7 +309,10 @@ class GithubAdapter extends GitAdapter {
   }
 
   protected getRepositoryUrl(repo: IRepo): string {
-    return `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`;
+    return (
+        `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`
+            .replace('api.', '')
+    );
   }
 
   private async checkActionSafety(repo: IRepo): Promise<SafetyStatus> {
--- a/src/services/github.ts
+++ b/src/services/github.ts
@@ -37,7 +37,7 @@ export default class GithubService {
 
       this.octokit = new RetryableThrottledOctokit({
         auth: token,
-        baseUrl: `https://${shepherdGitHubEnterpriseUrl}/api/v3`,
+        baseUrl: `https://${shepherdGitHubEnterpriseUrl}`,
         throttle: {
           enabled: false,
           onRateLimit: (retryAfter: number, options: any) => {


src/adapters/github.ts
```diff
@@ -309,7 +309,10 @@ class GithubAdapter extends GitAdapter {
   }
 
   protected getRepositoryUrl(repo: IRepo): string {
-    return `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`;
+    return (
+        `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`
```
aorinevo, line 313: Instead of replacing `api.`, I think we should just make the full url configurable instead of just the domain name.  I've added more details in a comment on issue #670

src/services/github.ts
```diff
@@ -37,7 +37,7 @@ export default class GithubService {
 
       this.octokit = new RetryableThrottledOctokit({
         auth: token,
-        baseUrl: `https://${shepherdGitHubEnterpriseUrl}/api/v3`,
+        baseUrl: `https://${shepherdGitHubEnterpriseUrl}`,
```
kavitha186, line 40: /api/v3 is required for GITHUB enterprise server apis, for the cloud the api/v3 is removed from the baseurl
https://docs.github.com/en/enterprise-cloud@latest/rest/quickstart?apiVersion=2022-11-28

taylorperkins: Tell me if I am missing something, but I do not see any `/api/v3` references in the docs that you provided. In the same way, I do not see a reference to `/api/v3` [anywhere in the enterprise docs](https://docs.github.com/en/enterprise-cloud@latest/search?query=api%2Fv3). In all of the examples in the quickstart you reference, GH uses examples like:
```
curl --request GET \
--url "https://api.github.com/repos/octocat/Spoon-Knife/issues" \
--header "Accept: application/vnd.github+json" \
--header "Authorization: Bearer YOUR-TOKEN"
```
Where the base URL does not have the `/api/v3` suffix.

aorinevo:
Interaction with GitHub Enterprise Server, GitHub Enterprise Cloud, or public GitHub via their APIs generally differ in URL as follows:
1. GitHub Enterprise Server: The API base URL is specific to the self-hosted instance (e.g., github.mycompany.com/api/v3)
2. GitHub Enterprise Cloud and Public GitHub: The API base URL is standard and publicly accessible via api.github.com

Given the above, what makes the most sense to me is to have `SHEPHERD_GITHUB_ENTERPRISE_URL` determine the full GitHub API vs just the base URL.  The default, for backwards compatibility, should point to api.github.com.  This also happens to coincide with (2).  

Thoughts?

src/adapters/github.ts
aorinevo, line 313: As this variable can hold either the GitHub public, Enterprise Cloud, or Enterprise Server URL, I'm thinking a more semantic variable name would be `githubUrl`.

aorinevo:
I've created the upstream branch `fix/github-enterprise-url` for this fix and updated your PR to point to it.

aorinevo:
Last note, could you please squash your commits into a single one and use our semantic commit message format?  We use the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

kavitha186:
@aorinevo, what will be the baseurl for enterprise cloud/public github?
I feel that setting up the github url along with baseurl as env var requires people who had  setup previously has to redo the enterprise url setup.

aorinevo:
The default would be `api.github.com/api/v3` which is what it was prior to the release of the GitHub Enterprise Server changes and hence, backwards compatible for any previous consumers of Shepherd.

aorinevo:
@taylorperkins, please let us know if you need help or are waiting on anything from our end to drive this forward.

aorinevo (approved):
Merging into feature branch.  Will make the outstanding changes directly on that branch.

aorinevo:
Thank you @taylorperkins for this contribution.  Will land this on main soon.

taylorperkins:
Thank you @aorinevo ! Sorry for not getting to the requested changes - I really haven't had the time recently. This is much appreciated! Cheers!

PR #683, opened 2024-01-31
fix(github-enterprise): configure gh enterprise full url via env var

aorinevo:


src/adapters/github.ts
```diff
@@ -309,7 +309,10 @@ class GithubAdapter extends GitAdapter {
   }
 
   protected getRepositoryUrl(repo: IRepo): string {
-    return `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`;
+    return (
+        `git@${shepherdGitHubEnterpriseUrl}:${repo.owner}/${repo.name}.git`
```
aorinevo, line 313: Drop `.replace` call.

aorinevo: Rename to gitHubEnterpriseUrl

aorinevo:
> Do we need to add an example in readme as well for setting the url for github enterprise server , cloud?

Yes, I was actually thinking of placing that in the env var section but those changes are on another feature branch.

kavitha186 (approved):

src/adapters/github.ts
aorinevo: This needs to be base url, not full url (e.g., api.github.com vs. api.github.com/api/v3)

aorinevo:
cc @taylorperkins

README.md
```diff
@@ -23,10 +23,11 @@ Install the Shepherd CLI:
 npm install -g @nerdwallet/shepherd
 ```
 
-If using GitHub Enterprise, ensure the following environment variable is exported:
+If using GitHub Enterprise, ensure the following environment variable are exported:
 
 ```
-export SHEPHERD_GITHUB_ENTERPRISE_URL={company_github_enterprise_url} # e.g., github.test.com
+export SHEPHERD_GITHUB_ENTERPRISE_URL={company_github_enterprise_base_url} # e.g., api.github.com
```
aorinevo, line 29: This should be SHEPHERD_GITHUB_ENTERPRISE_BASE_URL.

PR #693, opened 2024-02-04
feat(issues): create and list

kavitha186:
This PR is for the feature to post issues on the repositories with shepherd. if you are looking to post issues across multiple repositories, you can use the command "issue" to create, update , list posted issues and close issue

issue with the title, issue message, labels , state and state_reason provided in the migration scripts.

This enhancement feature is for the issue https://github.com/NerdWalletOSS/shepherd/issues/87 and https://github.com/NerdWalletOSS/shepherd/issues/590

### Create Issue

image

image

### Update Issue

image

### Closing Issue

image

image

### List Issues

image

## Screencasts

https://github.com/NerdWalletOSS/shepherd/assets/17735710/2f21596f-a943-4480-a1ce-dd0fdbfaf5f8


kavitha186:
image

kavitha186:
image

kavitha186:
Updating the issue:

image

kavitha186:
Closing the issue:

image


image

kavitha186:
List Posted issues

image

aorinevo:
@kavitha186, we need unit tests to cover list-issues and list commands.  Can you add those?

aorinevo (changes requested):

aorinevo:
@kavitha186, I've landed test coverage for all the existing commands.  You should be good to go to do the same for the new commands introduced in this PR.

src/services/github.ts
```diff
@@ -137,6 +140,32 @@ export default class GithubService {
...
   }
 
+  public createIssue(
+    criteria: RestEndpointMethodTypes['issues']['create']['parameters']
+  ): Promise<RestEndpointMethodTypes['issues']['create']['response']> {
+    return this.octokit.issues.create(criteria);
+  }
+
+  public updateIssue(criteria: {
+    owner: any;
```
aorinevo, line 150: I think we have a type we can type check criteria against.  If not can we convert this into a type and use it here?  Most of these are strings.  Can we be more restrictive on the type (e.g., owner is of type string instead of any)?

aorinevo:
@kavitha186, let’s add a screenshot of unit tests coverage.  A recording demoing the functionality would be helpful too.

kavitha186:
> @kavitha186, let’s add a screenshot of unit tests coverage. A recording demoing the functionality would be helpful too.

image

kavitha186:
> > @kavitha186, let’s add a screenshot of unit tests coverage. A recording demoing the functionality would be helpful too.
> 
> image


Uploading Screen-2024-03-17-185925.mp4…

kavitha186:
> @kavitha186, let’s add a screenshot of unit tests coverage. A recording demoing the functionality would be helpful too.


Uploading Screen-2024-03-17-185925.mp4…

src/commands/list-issues.test.ts
```diff
@@ -0,0 +1,78 @@
...
+                    owner: 'newOwner1',
+                    repo: 'newRepo1'
+                },
+            
+            ]
+            );
+
+            await listIssues(mockContext);
+
+            expect(process.stdout.write).toEqual(expect.any(Function))
```
aorinevo, line 75: process.stdout.write is generally a function, unless it's overwritten.  I imagine a snapshot test here would be effective as we would simply need to assert that the table rendered via process.stdout.write matches the snapshot.

aorinevo:
CI is failing due to linting errors.

kavitha186:
> CI is failing due to linting errors.

fixed

src/commands/issue.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    (getIssueListsFromTracker as jest.Mock).mockRejectedValueOnce([]);
+    await issue(mockContext);
+    expect(mockLogger.error).toHaveBeenCalledWith('Error to post/update issue');
+  });
+
+  it('should catch error when issue command is accessed', async () => {
+    await issue(mockContext1);
+    expect(mockSpinner.fail).toHaveBeenCalledWith('No issues in the shepherd yml to post');
+  });
+  it('should catch error when issue command is accessed', async () => {
```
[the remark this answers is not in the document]
kavitha186: fixed

src/commands/list-issues.test.ts
kavitha186: the snapshot for process.stdout doesn't seems to be working . I mocked the process.stdout function and validating if its being called In the command execution

kavitha186:
[

Uploading Screen-2024-03-17-185925.mp4…

](url)

kavitha186:
https://github.com/NerdWalletOSS/shepherd/assets/17735710/2f21596f-a943-4480-a1ce-dd0fdbfaf5f8

aorinevo:
I think we're in the final mile and very much looking forward to landing these changes.  There are a couple of things left to address:
1. There are a few type errors caught by github.  Can we look into that and resolve them?
2. When calling the list-issues command, we should include the state of the issue (e.g., closed, open, etc...).  Can we add that?

src/commands/list-issues.test.ts
aorinevo: Looks like this was resolved, however, the content of the table has values of undefined, apart from the column headings.  Should we expect to have non-null values here?

kavitha186:
> I think we're in the final mile and very much looking forward to landing these changes. There are a couple of things left to address:
> 
> 1. There are a few type errors to resolve which is causing the builds to fail.  Can we look into that and resolve them?
> 2. When calling the list-issues command, we should include the state of the issue (e.g., closed, open, etc...).  Can we add that?

@aorinevo I have a enum for state export enum state {
  open,
  closed,
}

src/commands/list-issues.ts
```diff
@@ -0,0 +1,19 @@
+import { IMigrationContext } from '../migration-context';
+import { getIssueTrackerFile } from '../util/persisted-data';
+import fs from 'fs-extra';
+import { table } from 'table';
+
+export default async (context: IMigrationContext) => {
+  const rows: any = [];
+
+  const columns = ['issue Number', 'issue Title', 'Owner', 'Repo Name'];
```
aorinevo, line 9: We want to include a status column.

TheSavior (commented):
Thanks for tagging me for feedback, I really appreciate the work you all have done here. I really hope this is  valuable for you too.

Something I'd really like is to be able to customize the description based on the repo. 

For example, say there are 5 issues I'm scanning repos for. If those were autofixable I'd fix them via PR, but they aren't, hence an issue.

So I want to check out the repos, run some scripts, and generate the issue description.

In my example, say I identify that the repo only has 3 of the 5 issues. The issue that I'd file might have some instructions for how to address those 3 issues, and even ideally be able to show the permalink to the lines in question: 

https://github.com/facebook/react-native/blob/38be5e35e6a8304d16e7eeac7bff1d4978384975/.babelrc#L5-L8

```
https://github.com/facebook/react-native/blob/38be5e35e6a8304d16e7eeac7bff1d4978384975/.babelrc#L5-L8
```

In order to piece that URL together I'd also need some information. I could probably get the sha by calling git log myself, and I can get the file path / line numbers. Not sure if I'd have the org/repo name though.

src/commands/list-issues.test.ts
kavitha186: yes passed the data and the table shows value now

README.md
```diff
@@ -82,6 +82,12 @@ hooks:
   post_checkout: npm install
   apply: mv .eslintrc .eslintrc.json
   pr_message: echo 'Hey! This PR renames `.eslintrc` to `.eslintrc.json`'
+issues:
```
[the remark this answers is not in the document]
kavitha186: If you see the list of commands in the ReadMe section, it has description for each command. 

- `pr`: Creates a PR for each repo with the message generated from the `pr_message` hook.
- `version`: Prints Shepherd version
- `issue`: Create, update, or close issues across multiple repository.

aorinevo:
@TheSavior and @kavitha186, thank you for your continued feedback and contributions.  While I think we've diverged a bit from the requirements Eli is after, I think Kavitha's foundational work is valuable and potentially adjustable to include Eli's use case.

**Use Case**
To align, let me summarize the use case we're after, leveraging Eli's example
0. Say there are 5 issues that we are scanning for in repos. If they are auto-fixable, we fix them via a PR.
1. For those issues which are not auto-fixable, we want to create an issue with a description that potentially contains data unique to that repo.

**Current State**
There are two checkout related hooks that are exposed via Shepherd's migration-spec file: `should_migrate` and `post_checkout`.  The `should_migrate` hook executes when a repo is checked out to determine whether to keep the repo.  If successful the repo is kept and subsequently the `post_checkout` hook is executed.  If `should_migrate` is unsuccessful, the repo is removed and `post_checkout` is not executed.

If `post_checkout` is successful, the repo is kept.  Otherwise, the repo is removed.

**Solution 1 (Recommended)**
Effort: Low

1. Leverage `should_migrate`
    a.  Instead of checking strictly whether or not the repo is auto-fixable for all 5 issues, we leverage this hook to generate the data to feed into the `post_checkout` step.  To make the data available between hooks, we can persist state to disk.
2. In the `post_checkout` step, we can read in the data and use it to take decisions on what issues to create and incorporate repo specific data into the issue details.

Pros: No code change required to Shepherd
Cons: End user has to persist and manage state on disk

**Solutions 2**
Effort: Medium

1. Similar to the recommended solution but we expose a singleton that can be used to pass state between hooks.

Pros: End user does not need to persist and manage state on disk
Cons: This state will only be accessible to hooks attached to the same command.

**Solution 3** 
Effort: High

1. Similar to Solution 2 but Shepherd persists and manages state on disk.

Pros: 
    - End user does not need to persist and manage state on disk.
    - End user can leverage state across hooks and commands.
Cons: 
    - We take a performance hit by reading/writing to disk.

README.md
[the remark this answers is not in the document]
aorinevo: @TheSavior, we're exposing issues as a top-level command in Shepherd (i.e., `shepherd issues ...`).  As it is currently implemented, it is mutually exclusive from other commands.

That said, I'm wondering if we should be thinking of incorporating hooks into issues.  For example, we can have a `should_create_issue` and `post_create_issue` hooks analogous to `should_migrate` and `post_checkout` hooks.

aorinevo:
@kavitha186, while I did some research on existing tooling to ensure we're not overlapping with feature sets already available elsewhere, I think it would be valuable to double check from your end.  

For example, does `gh` cli tool offer issue creation and management across multiple repos?

Just curious if there is anything we might have missed.

kavitha186:
> @kavitha186, while I did some research on existing tooling to ensure we're not overlapping with feature sets already available elsewhere, I think it would be valuable to double check from your end.  
> 
> For example, does `gh` cli tool offer issue creation and management across multiple repos?
> 
> Just curious if there is anything we might have missed.

I see there is a gh cli command to post issue. Create PRs. Perhaps that is available only to a single repo as far as I see.

TheSavior:
Option 1 seems fine tbh, happy to persist things to dish. It seems like there is some information I couldn't easily get from just a user script running in should_migrate though. For example, in order to form those URLs, I'd need to know the org/repo name. I could probably shell out to git remote and figure it out, would that be your recommendation? I would also probably need that to key the data stored on disk to the right repo so that when the issue hook runs later I know where to look for that persisted repo info.

TheSavior:
Oh or maybe your point about the gh cli is that I should just use shepherd to search and filter the repos, and once I get to should migrate and run the custom script, that script should just call the gh cli. That would probably work too!

aorinevo:
> Oh or maybe your point about the gh cli is that I should just use shepherd to search and filter the repos, and once I get to should migrate and run the custom script, that script should just call the gh cli. That would probably work too!

My comment with respect to adding the issues command to Shepherd.  That said, I think calling `gh` within your shepherd scripts would suffice for your use case.

kavitha186:
image

aorinevo:
@kavitha186, can we address the CI failures?

aorinevo:
@nwalters512, would love your input and feedback on this PR when you get a chance.

aorinevo:
Hi @nwalters512, when you get a chance, would love to get your feedback on this PR.

aorinevo:
@kavitha186, just a heads up that we're looking to land https://github.com/NerdWalletOSS/shepherd/pull/850 next.  It might make sense to point this PR towards the PR's branch or rebase off it so that you can get early signal on any additional changes needed to this PR.

aorinevo:
@kavitha186, I've created a v3 branch for the next major release of Shepherd.  Let's use that branch instead of main for these changes.

kavitha186:
@aorinevo I rebased the branch from main previously. Is there anything required to be taken care from my end?

aorinevo:
> @aorinevo I rebased the branch from main previously. Is there anything required to be taken care from my end?

Plan is to land https://github.com/NerdWalletOSS/shepherd/pull/852 and then this PR, both on v3 branch along with other features and enhancements.  Once landed, let's switch local use to v3 and let it bake for a certain period of time before we merge v3 down to main and release the next major version.  

What this means for this PR.  Once CI failures are resolved and #852 is landed, we can merge this PR into v3 provided the changes comport with changes in #852.  If not, it will likely require additional changes to this PR before landing.

aorinevo:
@kavitha186, when you get a chance, please resolve CI failures.

kavitha186:
> @kavitha186, when you get a chance, please resolve CI failures.

@aorinevo fixed the ci failures

aorinevo (approved):

PR #748, opened 2024-04-05
feat: Add https auth

jonathanmorley:
Shepherd currently uses ssh git urls to clone repositories.
This assumes that ssh auth is available in the environment for all repositories being cloned.

We are trying to use GitHub Actions to run Shepherd, with a GitHub App providing authentication.
This works for the API (we can have the GitHub App's API token get used by shepherd), but GitHub Apps do not have SSH keys associated with them to allow git pushes over ssh.

They can however use http authentication with their GitHub token to perform git pushes, which this PR allows for using (`SHEPHERD_GITHUB_ENTERPRISE_BASE_URL=x-access-token:${GITHUB_PAT}@github.com`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ npm install -g @nerdwallet/shepherd
 If using GitHub Enterprise, ensure the following environment variables are exported:
 
 ```
-export SHEPHERD_GITHUB_ENTERPRISE_BASE_URL={company_github_enterprise_base_url} # e.g., api.github.com
+export SHEPHERD_GITHUB_ENTERPRISE_BASE_URL={company_github_enterprise_base_url} # e.g., github.com
 export SHEPHERD_GITHUB_ENTERPRISE_URL={company_github_enterprise_url} # e.g., api.github.com/api/v3
 ```
 
--- a/src/adapters/github.ts
+++ b/src/adapters/github.ts
@@ -305,7 +305,7 @@ class GithubAdapter extends GitAdapter {
   }
 
   protected getRepositoryUrl(repo: IRepo): string {
-    const gitHubEnterpriseBaseUrl = process.env.SHEPHERD_GITHUB_ENTERPRISE_BASE_URL ?? 'api.github.com';
+    const gitHubEnterpriseBaseUrl = process.env.SHEPHERD_GITHUB_ENTERPRISE_BASE_URL ?? 'github.com';
     const githubProtocol = process.env.SHEPHERD_GITHUB_PROTOCOL ?? 'ssh';
 
     if (githubProtocol === 'ssh') {
--- a/src/adapters/github.ts
+++ b/src/adapters/github.ts
@@ -8,10 +8,6 @@ import { IEnvironmentVariables, IRepo } from './base';
 import GitAdapter from './git';
 import GithubService from '../services/github';
 
-const { SHEPHERD_GITHUB_ENTERPRISE_BASE_URL } = process.env;
-
-const gitHubEnterpriseBaseUrl = SHEPHERD_GITHUB_ENTERPRISE_BASE_URL || 'api.github.com';
-
 enum SafetyStatus {
   Success,
   PullRequestExisted,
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ npm install -g @nerdwallet/shepherd
 If using GitHub Enterprise, ensure the following environment variables are exported:
 
 ```
-export SHEPHERD_GITHUB_ENTERPRISE_BASE_URL={company_github_enterprise_base_url} # e.g., api.github.com
+export SHEPHERD_GITHUB_ENTERPRISE_BASE_URL={company_github_enterprise_base_url} # e.g., github.com
```
kavitha186, line 29: Should we add both as example Github enterprise server (github.com) and api.github.com (Cloud github enterprise)? @jonathanmorley

src/adapters/github.ts
```diff
@@ -309,7 +305,16 @@ class GithubAdapter extends GitAdapter {
   }
 
   protected getRepositoryUrl(repo: IRepo): string {
-    return `git@${gitHubEnterpriseBaseUrl}:${repo.owner}/${repo.name}.git`;
+    const gitHubEnterpriseBaseUrl = process.env.SHEPHERD_GITHUB_ENTERPRISE_BASE_URL ?? 'github.com';
```
kavitha186, line 308: How moving the gitHubEnterpriseBaseUrl initialized outside class?

README.md
jonathanmorley: Im not sure api.github.com is ever appropriate here.
I am using GitHub Enterprise Cloud, and github.com is the right value for this setting.
I would expect GitHub Enterprise Server to be installation-dependent.

src/adapters/github.ts
```diff
@@ -309,7 +305,16 @@ class GithubAdapter extends GitAdapter {
...
+    const githubProtocol = process.env.SHEPHERD_GITHUB_PROTOCOL ?? 'ssh';
```
aorinevo, line 309: Curious, what's the reasoning for defaulting to ssh instead of https?

README.md
aorinevo: When working with GitHub, the API endpoints differ based on whether you are using GitHub's public cloud service, GitHub Enterprise Cloud, or GitHub Enterprise Server (on-premises). Here’s how the endpoints vary:

* GitHub Public (GitHub.com)*: https://api.github.com/.
* GitHub Enterprise Cloud*: https://api.github.com/enterprise/[enterprise-name]
* GitHub Enterprise Server (On-Premises): https://[hostname]/api/v3/

In each case, the API functionality and how you interact with it are largely the same, but the base URL changes based on where your GitHub instance is hosted.

src/adapters/github.ts
aorinevo: As per my other comment, we should default to `api.github.com`.  Additionally, as this has already been rolled out to consumers, we would want to avoid breaking consumers that depend on it being defaulted to `api.github.com`.

aorinevo:
@jonathanmorley, thank you for contributing.  When you get a chance, can you update the PR description with context for these changes.

README.md
jonathanmorley: I think what you are describing is the API endpoint for GitHub, which is configurable in shepherd with `SHEPHERD_GITHUB_ENTERPRISE_URL`.

However, this environment variable is `SHEPHERD_GITHUB_ENTERPRISE_BASE_URL`, which appears to only be used as a hostname to generate git clone URLs.
I have observed both public github.com and github enterprise cloud using `github.com` for these values.

src/adapters/github.ts
jonathanmorley: The only option right now is ssh. Setting the default to ssh maintains the backwards compatability

README.md
cal4, line 1: Think it'd be useful to add a blurb to the Environment Variables table describing the new option, something like:
| `SHEPHERD_GITHUB_PROTOCOL`          | `ssh`                  | (`github` adapter) is the protocol to use when cloning repos. Can be `https` or `ssh`. `https` is useful when `ssh` is firewalled.                                                                                                                                                                                                                                                        |

README.md
cal4: I've observed this as well using Enterprise Cloud

cal4:
Thanks for contributing this!

src/adapters/github.ts
aorinevo: Hi @jonathanmorley, I've finally got some time to circle back on this.  Changes make sense and you're correct that as the code currently exists, it only supports `ssh`, though it's rather odd as I've certainly used it for `https` without making any changes.  

That said, the changes in this PR explicitly support https protocol and I'll work to land these changes this week.

At the moment I'm working through some dependency upgrades and ESM/CommonJS transpilation issues.  Once complete, this will be a fast follow.

aorinevo:
While CI failures are valid, landing on feature branch so that we can collaborate on upstream branch to resolve them.

aorinevo:
Please refer to https://github.com/NerdWalletOSS/shepherd/pull/847 to track progress towards landing on upstream.

PR #1019, opened 2025-07-25
fix(env-var): respect SHEPHERD_BASE_BRANCH

aorinevo:
This pull request introduces a new feature to the `GithubAdapter` class that allows the base branch to be determined dynamically based on an environment variable. It also includes corresponding unit tests to ensure the new behavior works as intended.

### Feature Enhancement:

* [`src/adapters/github.ts`](diffhunk://#diff-275c76f8995ad367ffe7a0c7f165d04a769a1e094ccf05a3b053ee1f71e2aa5bL293-R293): Updated the `getBaseBranch` method to check for the `SHEPHERD_BASE_BRANCH` environment variable. If it is set, the method returns its value; otherwise, it falls back to the repository's default branch (`repo.defaultBranch`).

### Testing:

* `src/adapters/github.test.ts`: Added a new test suite for the `getBaseBranch` method. This includes tests for scenarios where:
  - The `SHEPHERD_BASE_BRANCH` environment variable is not set, and the method defaults to the repository's default branch.
  - The `SHEPHERD_BASE_BRANCH` environment variable is set, and the method returns its value.
  - The `SHEPHERD_BASE_BRANCH` environment variable is an empty string, and the method returns an empty string.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,12 +87,10 @@ hooks:
 ### Fields
 
 - `id`:
-
   - **Description**: Specifies a unique identifier for this migration.
   - **Usage**: Used as a branch name for the migration and internally by Shepherd to track migration state.
 
 - `title`:
-
   - **Description**: A title for the migration.
   - **Usage**: Used as the commit message.
 
@@ -109,17 +107,14 @@ hooks:
 Hooks define the core functionality of a migration in Shepherd.
 
 - `should_migrate`:
-
   - **Description**: Commands to determine if a repo requires migration.
   - **Behavior**: Non-zero exit values indicate the repo should not be migrated.
 
 - `post_checkout`:
-
   - **Description**: Commands executed after a repo passes `should_migrate` checks.
   - **Usage**: Ideal for one-time setup actions per repo, like installing dependencies.
 
 - `apply`:
-
   - **Description**: Commands that perform the actual migration.
   - **Note**: This can range from simple to complex sequences, depending on migration needs.
 
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 <img alt="Illustration of a sheep" width=160 align=right src="images/shepherd-logo.png">
 
+[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](./CONTRIBUTING.md)
 [![GitHub Workflow Status](https://img.shields.io/github/actions/workflow/status/NerdWalletOSS/shepherd/release.yml?style=flat-square)](https://github.com/NerdWalletOSS/shepherd/actions)
 [![semantic-release: conventionalcommits](https://img.shields.io/badge/semantic--release-conventionalcommits-e10079?logo=semantic-release)](https://github.com/semantic-release/semantic-release)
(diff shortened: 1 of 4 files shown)


Siddhi-Iccha:
Hi @aorinevo 👋
I see we made same bug fix for the same issue. Here is my [PR - #1021](https://github.com/NerdWalletOSS/shepherd/pull/1021) which also fixes one brittle test along with addition of contribution badge.
I am fine with either both of these fixes the issue.

Siddhi-Iccha:
> @Siddhi-Iccha, I would like to incorporate your changes onto this PR. Would you be open to opening a PR against this branch?

Sure, [here](https://github.com/NerdWalletOSS/shepherd/pull/1021) it is

aorinevo:
@kavitha186, when you get a chance, please review.

kavitha186:
@aorinevo 

I got different thought on this feature. Please let me know your feedback on the same.

1. Should we introduce a --base-branch argument?

Explicit --base-branch argument in both the PR creation and checkout commands.

Why?

Relying on an environment variable (like BASE_BRANCH) might lead to forcefully use it for all changes and it is not visible to user

If it's not set or set incorrectly, PRs might be raised against the wrong base (e.g., main instead of release/xyz), which could cause confusion or unintended PRs.

Proposed behavior:

base-branch can be considered for the PR creation

An alternative approach if we are going to go with env variable, If we are going to rely on env variable base branch , show a warning like: ⚠️ Warning: Base branch is set in environment variable. Falling back to base branch as per env variable instead of main.

2. Should base branch also be used during git checkout?


If you checkout a feature branch from main, but the PR is intended to be against another branch like release/1.2, then:

Your diff will include unrelated commits.

Git conflicts during the merge can arise later.

Reviewers will see misleading changes.

Proposed improvement:

 Create a branch off from base branch

Summary of Changes We Might Add

Accept --base-branch arg in both PR creation and checkout commands.

Use it to: 

Determine the correct base for PR.

Check out the feature branch from that base branch.

Show a warning when no base branch is specified explicitly.

aorinevo:
> 1. Should we introduce a --base-branch argument? 

I think we would want to expose a top-level cli command to provide more controls.  That said, I don't think this is isolated to base branch.  I'm wondering what other env vars are set that could benefit from this and should we allow for override via options?  

> 2. Should base branch also be used during git checkout?

Shepherd first attempts to checkout `this.branchName` from origin using SimpleGit.  If that fails, it checks out a local branch off of upstream default branch.  Finally, if that fails, it means the branch already exists locally and it simply switches to that branch.

```js
    try {
      await this.git(repo).checkout(['-b', this.branchName, `origin/${this.branchName}`]);
    } catch {
      try {
        await this.git(repo).checkoutLocalBranch(this.branchName);
      } catch {
        // This branch probably already exists; we'll just switch to it
        // to make sure we're on the right branch for the commit phase
        await this.git(repo).checkout(this.branchName);
      }
    }
```

For branching strategies where there are long-lived branches, this certainly makes sense.  

## Next Steps

How we want to approach (1) and (2) is non-blocking with respect to the changes in this PR.  I recommend landing these changes and opening an issue to sort out how we want to proceed with solving for (1) and (2).

@kavitha186, thoughts?

PR #1065, opened 2025-10-19
feat!: drop node v18 support, add node v24

aorinevo:
This pull request updates the Node.js version requirements and dependencies to ensure compatibility with newer Node.js releases. The most important changes include updating the supported Node.js versions for CI and release workflows, specifying a minimum Node.js version in `package.json`, and upgrading key dependencies.

**Node.js version updates:**

* CI workflow now tests against Node.js versions 20, 22, and 24, dropping support for version 18 (`.github/workflows/ci.yml`).
* Release workflow now uses Node.js version 22 instead of 20 (`.github/workflows/release.yml`).
* The minimum required Node.js version is now specified as 20 or higher in the `engines` field of `package.json`.

**Dependency updates:**

* Upgraded the `ora` package from version 8.1.0 to 9.0.0 in `package.json`.
* Fixed the order of dependencies in `package.json` so that `table` is correctly listed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: ['18', '20', '22']
+        node: ['20', '22', '24']
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


aorinevo:
Closes https://github.com/shepherd-tools/shepherd/issues/947

